PR #5, opened 2020-04-16
6161: Unauthorized error page.

BenBJSS:


app/pages/error/contextCreator.test.js
```diff
@@ -0,0 +1,82 @@
...
+    const context = getContext({
+      description: 'Error Description',
+    });
+
+    expect(context.error.status).toEqual(defaultContext.error.status);
+    expect(context.error.backLinkText).toEqual(defaultContext.error.backLinkText);
+    expect(context.error.backLinkHref).toEqual(defaultContext.error.backLinkHref);
+    expect(context.error.title).toEqual(defaultContext.error.title);
+    expect(context.error.description).toEqual(context.error.description);
+  });
```
alibombaye, line 81: Nice but I would just assert the single thing you are testing for. So here just the description and the rest can be removed. The first test ensures all the defaults are checked

PR #13, opened 2020-04-29
6101 update with claim type

alibombaye:
[AB#6440](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/6440)

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "axios": "^0.19.0",
     "body-parser": "^1.19.0",
     "buying-catalogue-components": "^1.1.42",
-    "buying-catalogue-library": "^1.0.5",
+    "buying-catalogue-library": "^1.0.6",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -33,7 +33,7 @@
     "axios": "^0.19.0",
     "body-parser": "^1.19.0",
     "buying-catalogue-components": "^1.1.56",
-    "buying-catalogue-library": "^1.0.5",
+    "buying-catalogue-library": "^1.0.6",
```
kat1906, line 36: what does this bring in?

alibombaye: I don't know being honest. I just updated it to the latest since I was having some issues with getting the Unauthenticated page to show. But it was something else that was causing that to break.

PR #16, opened 2020-04-29
Refactor start redis node script to use docker-compose

assoomro:
[AB#6273](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/6273)


assoomro:
> Nice and will this check if redis is already running and not start a new one?

Yeah, it checks whether anything has been changed and either recreates the container or leaves it be.

image

assoomro:
> Fab you've removed the first ever bit of shell scripting I've ever done 👍

😂 It was great while it lasted

PR #20, opened 2020-05-01
added healthchecks from bcl

alibombaye:
[AB#6538](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/6538)


app/routes.js
```diff
@@ -15,6 +15,8 @@ const addContext = ({ context, user, csrfToken }) => ({
 export const routes = (authProvider) => {
   const router = express.Router();
 
+  healthRoutes({ router, dependencies: getHealthCheckDependencies(config), logger });
```
kat1906, line 18: what are you getting from the config? how does the healthcheck routes know how unhealthy or degraded dependencies affects it? for this app it will be unhealthy if either isapi or ordapi are down but what about pb when it will be unhealthy if bapi is down but not if dapi or isapi are down?

alibombaye: There is flag that is passed in as part of dependencies that will tell the healthRoutes if that dependency is critical or not. If one of these critical dependencies fails it is `unhealthy`. If it isn't then the ready status is `degraded`

kat1906: ah yeah I was misunderstanding soz

PR #23, opened 2020-05-05
Feature/6353 authprovider bcl

alibombaye:
[AB#6698](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/6698)


app/helpers/routerHelper.js
```diff
@@ -6,13 +6,15 @@ export const withCatch = (authProvider, route) => async (req, res, next) => {
   try {
     return await route(req, res, next);
   } catch (err) {
+    if (err instanceof ErrorContext) {
```
kat1906, line 9: what is the significance of the order of these `if`s in this file?

alibombaye: Just thought we should handle this one first.

kat1906: fair 😄

PR #27, opened 2020-05-06
4826: Add dashboard page with header and footer

kat1906:
Task: [AB#6752](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/6752)


app/config.js
```diff
@@ -17,6 +17,9 @@ module.exports = {
   // The base uri the app is running on to pass to identity service for redirection
   appBaseUri: process.env.APP_BASE_URI || 'http://localhost:3006',
 
+  // BLOBSTORE_HOST
+  blobstoreHost: process.env.BLOBSTORE_HOST || 'https://gpitfuturesdevsa.blob.core.windows.net',
```
alibombaye, line 21: why do we need this?

alibombaye: ah ok ignore me I can see why

app/config.js
```diff
@@ -42,7 +45,7 @@ module.exports = {
   redisUrl: process.env.REDIS_URL || 'localhost',
 
   // The port redis is running
-  redisPort: process.env.REDIS_PORT || 6379,
+  redisPort: process.env.REDIS_PORT || 6380,
```
alibombaye, line 48: oh yeaahhh!!!! makes me wonder how I'm getting it to work for me locally via the cluster

kat1906: I think the cluster one is 6379 and local is 6380 so I will change it back

alibombaye: yeah that's right. got confused

PR #29, opened 2020-05-06
4826: Create dashboard page

kat1906:
Task: [AB#6754](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/6754)

### DIFF
--- a/app/config.js
+++ b/app/config.js
@@ -6,7 +6,7 @@ module.exports = {
   baseUrl: process.env.BASE_URL || '/order',
 
   // Public browse base URL
-  publicBrowseBaseUrl: process.env.PUBLIC_BROWSE_BASE_URL || '/',
+  publicBrowseBaseUrl: process.env.PUBLIC_BROWSE_BASE_URL || 'http://localhost:3000',
 
   // Environment
   env: process.env.NODE_ENV || 'development',
@@ -45,7 +45,7 @@ module.exports = {
   redisUrl: process.env.REDIS_URL || 'localhost',
 
   // The port redis is running
-  redisPort: process.env.REDIS_PORT || 6379,
+  redisPort: process.env.REDIS_PORT || 6380,
 
   // Boolean to indicate whether to connect to redis via TLS
   redisTls: process.env.REDIS_TLS || 'false',
--- a/app/config.js
+++ b/app/config.js
@@ -45,7 +45,7 @@ module.exports = {
   redisUrl: process.env.REDIS_URL || 'localhost',
 
   // The port redis is running
-  redisPort: process.env.REDIS_PORT || 6380,
+  redisPort: process.env.REDIS_PORT || 6379,
 
   // Boolean to indicate whether to connect to redis via TLS
   redisTls: process.env.REDIS_TLS || 'false',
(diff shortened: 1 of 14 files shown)


app/config.js
```diff
@@ -45,7 +45,7 @@ module.exports = {
   redisUrl: process.env.REDIS_URL || 'localhost',
 
   // The port redis is running
-  redisPort: process.env.REDIS_PORT || 6379,
+  redisPort: process.env.REDIS_PORT || 6380,
```
alibombaye, line 48: change this back pretty please

app/pages/dashboard/contextCreator.js
```diff
@@ -1,5 +1,6 @@
 import manifest from './manifest.json';
 
-export const getContext = () => ({
+export const getContext = ({ orgId }) => ({
   ...manifest,
+  title: `${orgId} orders`,
```
alibombaye, line 5: In admin we had that fancy function. Could we use that here? or even put that fancy function in BCL

app/pages/dashboard/manifest.json
```diff
@@ -1,3 +1,28 @@
 {
-  "title": "Dashboard Page"
+  "backlinkText": "Return to homepage",
+  "description": "Manage orders currently in progress and view orders that have been submitted.",
+  "newOrderButtonText": "Create new order",
+  "newOrderButtonHref": "#",
```
alibombaye, line 5: We do have this route now. `/organisation/neworder`

app/styles/main.scss
```diff
@@ -3,3 +3,6 @@
 
 // Buying catalogue library styles
 @import 'node_modules/buying-catalogue-components/app/styles/main';
+
+// Pages
+@import '../pages//dashboard/dashboard';
```
alibombaye, line 8: double `/` here

app/config.js
alibombaye: By the way you can set this in your .env file. Save you from changing it back and forth when you run it locally

app/test-utils/helper.js
```diff
@@ -19,3 +19,8 @@ export const getCsrfTokenFromGet = async (app, getPath, authenticationCookie) =>
     });
   return { cookies, csrfToken };
 };
+
+export const extractInnerText = async (element) => {
+  const elementInnerText = await element.innerText;
+  return elementInnerText.trim();
+};
```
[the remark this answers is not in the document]
kat1906: As this is the last all I feel like there is no point now for things like this as they are pretty stable and it is more work moving them to BCL then bumping all the apps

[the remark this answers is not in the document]
alibombaye: yeah fair enough. I'll add to techDeb. At the rate we are flying through these tasks we'll be fighting for work towards the end of this sprint

app/pages/dashboard/contextCreator.js
kat1906: the reason I didn't want to do this for the title is because of the way it shows in the <title> tag in the page tab in the browser

PR #32, opened 2020-05-07
Feature/5111 disable button

alibombaye:
Disabled buttons shown on create order page

Screenshot 2020-05-07 at 18 01 41

[AB#6765](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/6765)

### DIFF
--- a/app/pages/order-task-list/template.njk
+++ b/app/pages/order-task-list/template.njk
@@ -23,6 +23,7 @@
         {{ bcButton({
           dataTestId: "delete-order-button",
           text: deleteOrderButtonText,
+          altText: deleteOrderButtonAltText,
           classes: "nhsuk-button--secondary nhsuk-u-margin-bottom-9",
           href: deleteOrderButtonHref,
           enabled: deleteOrderButtonEnabled
--- a/app/pages/order-task-list/template.njk
+++ b/app/pages/order-task-list/template.njk
@@ -36,6 +36,7 @@
             {{ bcButton({
               dataTestId: "preview-order-button",
               text: previewOrderButtonText,
+              altText: previewOrderButtonAltText,
               classes: "nhsuk-button--secondary nhsuk-u-margin-bottom-9 nhsuk-u-margin-right-5",
               href: previewOrderButtonHref,
               enabled: previewOrderButtonEnabled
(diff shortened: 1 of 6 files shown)


kat1906:
there needs to be an alt property on the button when it is disabled for screen readers:
- The preview order button will be disabled until you complete and save step 1
- The delete order button will be disabled until you complete and save step 1
- The submit order button will be disabled until each step is complete
(content from Luke)

alibombaye:
The NHS button component does add an aria-label when the button is disabled
https://github.com/nhsuk/nhsuk-frontend/tree/master/packages/components/button

```
<button class="nhsuk-button nhsuk-button--disabled" type="submit" disabled="disabled" aria-disabled="true">
  Disabled button
</button>
```
This is the rendered HTML for the button component. Think this should cover the above ACs?

kat1906:
I think it is supposed to be those sentences on the `alt` key. Maybe you would be better speaking to Luke?

kat1906:
Do you need to amend the button component in BCC?

alibombaye:
Screenshot 2020-05-11 at 11 29 28
Alt text added - @kat1906 can you take another look please

app/pages/order-task-list/template.njk
```diff
@@ -16,5 +17,44 @@
         <h2 data-test-id="{{pageName}}-page-description" class="nhsuk-u-font-size-24 nhsuk-u-margin-bottom-6">{{ description }}</h2>
       </div>
     </div>
+
+    <div class="nhsuk-grid-row">
+      <div class="nhsuk-grid-column-one-third">
+        {{ bcButton({
+          dataTestId: "delete-order-button",
+          text: deleteOrderButtonText,
+          altText: deleteOrderButtonAltText,
```
kat1906, line 26: this alt text is only for when it is disabled, does it need conditionally added in another task?

alibombaye: yes, when we come to enable it then we'll need to change it to show something else. Even if it just says the name of the button.

kat1906: can you write a note on the relevant ticket as I can envisage this being forgotten / missed

alibombaye: added a task for the save order story regarding when some of these buttons are enabled

PR #34, opened 2020-05-11
5322: Add description page

kat1906:
Story: [AB#5322](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/5322)
Task: [AB#6774](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/6774)
Screenshot 2020-05-11 at 11 29 23

### DIFF
--- a/app/pages/items/description/template.njk
+++ b/app/pages/items/description/template.njk
@@ -1,6 +1,6 @@
 {% extends 'includes/layout.njk' %}
 {% from 'components/back-link/macro.njk' import backLink %}
-{% from 'components/button/macro.njk' import button %}
+{% from 'components/bc-button/macro.njk' import bcButton %}
 {% from 'components/textarea-field-no-count/macro.njk' import textareaFieldNoCount %}
 
 {% block body %}
@@ -20,7 +20,7 @@
         {{ textareaFieldNoCount(descriptionQuestion) }}
 
         <div data-test-id="save-button">
-          {{ button({
+          {{ bcButton({
               text: saveButtonText,
               href: saveButtonHref,
               classes: "nhsuk-u-margin-top-9"
--- a/app/pages/items/description/template.njk
+++ b/app/pages/items/description/template.njk
@@ -1,0 +1,31 @@
+{% extends 'includes/layout.njk' %}
+{% from 'components/back-link/macro.njk' import backLink %}
+{% from 'components/button/macro.njk' import button %}
+{% from 'components/textarea-field-no-count/macro.njk' import textareaFieldNoCount %}
+
+{% block body %}
+  <div class="nhsuk-width-container" data-test-id="description-page">
+    <div data-test-id="go-back-link">
+      {{ backLink({
+        "href": backlinkHref,
+        "text": backlinkText
+      }) }}
+    </div>
+
+    <div class="nhsuk-grid-row">
+      <div class="nhsuk-grid-column-two-thirds">
+        <h1 data-test-id="description-page-title" class="nhsuk-u-margin-top-7 nhsuk-u-margin-bottom-3">{{ title }}</h1>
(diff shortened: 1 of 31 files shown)


app/pages/items/description/template.njk
```diff
@@ -0,0 +1,31 @@
...
+
+    <div class="nhsuk-grid-row">
+      <div class="nhsuk-grid-column-two-thirds">
+        <h1 data-test-id="description-page-title" class="nhsuk-u-margin-top-7 nhsuk-u-margin-bottom-3">{{ title }}</h1>
+        <h2 data-test-id="description-page-description" class="nhsuk-u-font-size-24 nhsuk-u-margin-bottom-6">{{ description }}</h2>
+
+        {{ textareaFieldNoCount(descriptionQuestion) }}
+
+        <div data-test-id="save-button">
+          {{ button({
```
alibombaye, line 23: can we use the new bcButton component. I just created a PR so the altText will be default to the button text so should work here

app/pages/items/description/template.test.js
```diff
@@ -0,0 +1,58 @@
...
+
+  it('should render the textareaFieldNoCount component', componentTester(setup, (harness) => {
+    harness.request(context, ($) => {
+      const input = $('[data-test-id="question-description"] textarea');
+      const footerAdvice = $('[data-test-id="textarea-field-footer"] span');
+
+      expect(input.length).toEqual(1);
+      expect(input[0].attribs.id).toEqual(context.descriptionQuestion.question.id);
+      expect(input[0].attribs.name).toEqual(context.descriptionQuestion.question.id);
+      expect(input[0].attribs.rows.trim())
```
alibombaye, line 43: why the `[0]` is the selector finding an array? should it just be finding the single element

kat1906: it doesn't work without `[0]`

app/pages/items/description/template.njk
kat1906: it is disabled?

app/pages/items/description/template.test.js
alibombaye: for the selector can you try this;
`const input = $('[data-test-id="question-description"]').find('textarea')`;

kat1906: it is syntactic sugar for what I have written in line 37

kat1906: I think it is fine as the length is still getting checked - splitting hairs here

app/pages/items/description/template.njk
kat1906: see comment on slack

alibombaye: Nope but we can pass in the enabled flag. It will just provide a nice little uniform way we handle all our buttons in OF going forward.

app/pages/items/description/template.test.js
alibombaye: just looks confusing that's all. If something we expect to be a single element we should reflect that. The data-test-id should help us getting that single element

app/pages/items/description/template.njk
```diff
@@ -0,0 +1,31 @@
...
+    </div>
+
+    <div class="nhsuk-grid-row">
+      <div class="nhsuk-grid-column-two-thirds">
+        <h1 data-test-id="description-page-title" class="nhsuk-u-margin-top-7 nhsuk-u-margin-bottom-3">{{ title }}</h1>
+        <h2 data-test-id="description-page-description" class="nhsuk-u-font-size-24 nhsuk-u-margin-bottom-6">{{ description }}</h2>
+
+        {{ textareaFieldNoCount(descriptionQuestion) }}
+
+        <div data-test-id="save-button">
```
alibombaye, line 22: you can pass this dataTestId directly to the component and we don't need the this div then

PR #38, opened 2020-05-12
4826: Update endpoint for dashboard

kat1906:
Task: [AB#6884](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/6884)

### DIFF
--- a/app/pages/dashboard/ui.test.js
+++ b/app/pages/dashboard/ui.test.js
@@ -17,7 +17,7 @@ const setCookies = ClientFunction(() => {
 
 const mocks = () => {
   nock(orderApiUrl)
-    .get('/api/v1/orders')
+    .get('/api/v1/organisation/org1/orders')
     .reply(200, mockOrdersData);
 };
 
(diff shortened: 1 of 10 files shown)


app/pages/dashboard/ui.test.js
```diff
@@ -17,7 +17,7 @@ const setCookies = ClientFunction(() => {
 
 const mocks = () => {
   nock(orderApiUrl)
-    .get('/api/v1/orders')
+    .get('/api/v1/organisation/undefined/orders')
```
alibombaye, line 20: we should be able to pass in an orgId from the claim right? might need to update that fake cookie we set up to return a primaryOrganisationId

kat1906: I have no idea what to do with this, tech debt?

testcafeRunner.js
```diff
@@ -17,7 +17,7 @@ const browserToRun = browserFromArgs.length > 0 ? browserFromArgs : 'chrome:head
 const testFromArgs = process.argv.slice(3, 4);
 const testsToRun = testFromArgs ? `**/*${testFromArgs}*/ui.test.js` : '**/*ui.test.js';
 
-let concurrency = 4;
+let concurrency = 1;
```
alibombaye, line 20: Next PR whoever gets in there first. Lets take this out. The concurrency doesn't work with the nock checks

kat1906: take out completely?

PR #39, opened 2020-05-12
Feature/5321 show task list

alibombaye:
Screenshot 2020-05-12 at 14 20 19

[AB#6772](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/6772)

### DIFF
--- a/app/pages/order-task-list/contextCreator.js
+++ b/app/pages/order-task-list/contextCreator.js
@@ -4,9 +4,9 @@ import { generateTaskList } from './helpers/generateTaskList';
 import { baseUrl } from '../../config';
 
 
-export const getContext = ({ pageName }) => ({
+export const getContext = ({ orderId }) => ({
   ...neworderPageManifest,
-  taskList: generateTaskList({ orderId: pageName, taskListManifest }),
+  taskList: generateTaskList({ orderId, taskListManifest }),
   backLinkHref: `${baseUrl}/organisation`,
-  pageName,
+  orderId,
 });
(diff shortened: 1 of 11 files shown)


app/pages/order-task-list/contextCreator.js
```diff
@@ -1,8 +1,12 @@
 import neworderPageManifest from './neworder/manifest.json';
+import taskListManifest from './taskListManifest.json';
```
kat1906, line 2: should this just be manifest in line with all the rest?

alibombaye: We are going to need it for the existing order task-list page. So that's why I took it out here to its own

kat1906: ok

app/pages/order-task-list/contextCreator.js
```diff
@@ -1,8 +1,12 @@
...
+import { generateTaskList } from './helpers/generateTaskList';
 import { baseUrl } from '../../config';
 
+
 export const getContext = ({ pageName }) => ({
   ...neworderPageManifest,
+  taskList: generateTaskList({ orderId: pageName, taskListManifest }),
```
kat1906, line 9: instead of pageName I just used orderId throughout and the id of a new order is 'neworder'

alibombaye: 👍

kat1906:
also might need a bit more space below the page description

PR #41, opened 2020-05-12
4826: Show org name on dashboard page

BenBJSS:


kat1906:
please can you add to the ui tests for dashboard as at the moment they look for "undefined order" as the title

PR #43, opened 2020-05-13
Feature/5161 existing order title description

BenBJSS:


app/endpoints.js
```diff
@@ -3,6 +3,7 @@ import { oidcBaseUri, orderApiUrl } from './config';
 const endpoints = {
   getIdentityApiHealth: () => `${oidcBaseUri}/health/ready`,
   getOrders: options => `${orderApiUrl}/api/v1/organisation/${options.orgId}/orders`,
+  getExistingOrder: options => `${orderApiUrl}/api/v1/orders/${options.orderId}/summary`,
```
kat1906, line 6: I think this should be changed to `gretOrderSummary`

app/pages/order-task-list/contextCreator.js
```diff
@@ -1,12 +1,28 @@
...
   ...neworderPageManifest,
   taskList: generateTaskList({ orderId, taskListManifest }),
   backLinkHref: `${baseUrl}/organisation`,
   orderId,
 });
+
+export const getExistingOrderContext = ({ orderId, orderDescription }) => ({
+  ...existingorderPageManifest,
+  title: `${existingorderPageManifest.title} ${orderId}`,
+  orderDescription,
```
kat1906, line 17: what is this for? you haven't passed description into any of your tests so is it needed?

BenBJSS: We have, line 55 contextCreator.test.js

alibombaye: this is to show the description we get from the endpoint. We pass it in from the controller. There is a test for it in the contextCreator.test.js

app/pages/order-task-list/contextCreator.js
```diff
@@ -1,12 +1,28 @@
...
+  taskList: generateTaskList({ orderId, taskListManifest }),
+  backLinkHref: `${baseUrl}/organisation`,
+  orderId,
+});
+
+export const getContext = ({ orderId, orderDescription }) => {
+  if (orderId === 'neworder') {
```
kat1906, line 24: a ternary might make this neater but not essential

BenBJSS: We had it like that, linting complained

alibombaye: ok the ternary option looks like this. Think the if version is bit more readable
```
export const getContext = ({ orderId, orderDescription }) => (orderId === 'neworder'
  ? getNewOrderContext({ orderId })
  : getExistingOrderContext({ orderId, orderDescription }));
```

app/pages/order-task-list/controller.js
```diff
@@ -1,3 +1,20 @@
...
+import { getEndpoint } from '../../endpoints';
+import { logger } from '../../logger';
 
-export const getNewOrderPageContext = () => getContext({ orderId: 'neworder' });
+const getNewOrderTaskListPageContext = ({ orderId }) => getContext({ orderId });
+
+const getExistingOrderTaskListPageContext = async ({ accessToken, orderId }) => {
+  const endpoint = getEndpoint({ endpointLocator: 'getOrderSummary', options: { orderId } });
+  const order = await getData({ endpoint, accessToken, logger });
+  logger.info('Existing order returned');
```
kat1906, line 11: could add the id to the log

app/pages/order-task-list/controller.js
```diff
@@ -1,3 +1,20 @@
...
+  return getContext({ orderId, orderDescription: order.description });
```
kat1906, line 12: can we call it description?

app/pages/order-task-list/existingorder/manifest.json
```diff
@@ -0,0 +1,24 @@
+{
+  "backLinkText": "Go back",
```
kat1906, line 2: is this manifest the same as the other one?

kat1906: maybe we could use the same one and move the differing bits into the contextCreator instead of 2 different manifests. It depends how much is different

app/pages/order-task-list/controller.js
alibombaye: could get a little confusing since we have a description for the description of the page too. Think it might be better to be explicit here.

app/routes.js
```diff
@@ -42,14 +42,11 @@ export const routes = (authProvider) => {
...
-  router.get('/organisation/neworder', authProvider.authorise({ claim: 'ordering' }), withCatch(authProvider, async (req, res) => {
-    const context = getNewOrderPageContext();
-    res.render('pages/order-task-list/template.njk', addContext({ context, user: req.user }));
-  }));
-
   router.get('/organisation/:orderId', authProvider.authorise({ claim: 'ordering' }), withCatch(authProvider, async (req, res) => {
+    const accessToken = extractAccessToken({ req, tokenType: 'access' });
     const { orderId } = req.params;
-    res.status(200).send(`existing order ${orderId} page`);
+    const context = await getOrderTaskListPageContext({ accessToken, orderId });
```
kat1906, line 48: also don't like order in this name - it is the order form about orders and there is only 1 task list

app/pages/order-task-list/existingorder/manifest.json
alibombaye: I think only the backLinkText is the same as what we have in neworder manifest. Think it might be better to just duplicate this for now

app/routes.test.js
```diff
@@ -129,37 +130,40 @@ describe('routes', () => {
       }));
   });
 
-  describe('GET /organisation/neworder', () => {
-    const path = '/organisation/neworder';
+  describe('GET /organisation/:orderId', () => {
+    const path = '/organisation/order-id';
```
kat1906, line 134: is this path correct?

kat1906: below you are talking about neworder

app/routes.js
alibombaye: ok yeah fair enough. I'll update that

app/pages/order-task-list/contextCreator.js
kat1906: although personally I like this better

app/pages/order-task-list/controller.js
kat1906: ok

app/pages/order-task-list/existingorder/manifest.json
kat1906: really? all the button stuff looks the same

kat1906: dunno whether it is worth having a common manifest for the common stuff otherwise there are effectively 2 points of truth for the same page

app/routes.test.js
alibombaye: yes. I believe so. What do you think is wrong with it?

kat1906: in line 142 you are returning the orderId value of 'neworder' but this path says 'order-id' as orderId

alibombaye: ok I'm with you now. Yes you are right 👍

app/pages/order-task-list/existingorder/manifest.json
alibombaye: The buttons will be different in terms of there paths and whether they are enabled or not. The names of the button will be the same. We'll take a look at this though when we get round to the button part of this story. Good spot

kat1906: we might need a function that populates the buttons with the info they need in the end

PR #47, opened 2020-05-14
BUG: Amend getOrders url

kat1906:


kat1906:
actually I will probably need to fix some tests too brb

PR #64, opened 2020-05-20
4615: Call oapi for org data

kat1906:
Story: [AB#4615](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/4615)
Task [AB#7086](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/7086)

### DIFF
--- a/app/pages/sections/call-off-ordering-party/controller.js
+++ b/app/pages/sections/call-off-ordering-party/controller.js
@@ -25,4 +25,5 @@ export const getCallOffOrderingPartyContext = async ({ orderId, orgId, accessTok
       throw new Error();
     }
   }
+  throw new Error();
 };
--- a/app/pages/sections/call-off-ordering-party/controller.js
+++ b/app/pages/sections/call-off-ordering-party/controller.js
@@ -20,7 +20,6 @@ export const getCallOffOrderingPartyContext = async ({ orderId, orgId, accessTok
     try {
       const orgDataEndpoint = getEndpoint({ endpointLocator: 'getOrganisationById', options: { orgId } });
       const orgData = await getData({ endpoint: orgDataEndpoint, accessToken, logger });
-      // const orgData = mockOrgData;
       logger.info(`Organisation with id: ${orgId} found`);
       return getContext({ orderId, data: orgData });
     } catch (error) {
--- a/app/pages/sections/call-off-ordering-party/controller.js
+++ b/app/pages/sections/call-off-ordering-party/controller.js
@@ -3,14 +3,29 @@ import { getContext } from './contextCreator';
 import { getEndpoint } from '../../../endpoints';
 import { logger } from '../../../logger';
 
-export const getCallOffOrderingPartyContext = async ({ orderId, accessToken }) => {
-  const callOffOrgDataEndpoint = getEndpoint({ endpointLocator: 'getCallOffOrderingParty', options: { orderId } });
-  const callOffOrgData = await getData({ endpoint: callOffOrgDataEndpoint, accessToken, logger });
+export const getCallOffOrderingPartyContext = async ({ orderId, orgId, accessToken }) => {
+  try {
+    const callOffOrgDataEndpoint = getEndpoint({ endpointLocator: 'getCallOffOrderingParty', options: { orderId } });
+    const callOffOrgData = await getData({ endpoint: callOffOrgDataEndpoint, accessToken, logger });
 
-  if (callOffOrgData) {
-    logger.info(`Call off ordering party found in ORDAPI for ${orderId}: ${callOffOrgData}`);
-    return getContext({ orderId, data: callOffOrgData });
+    if (callOffOrgData) {
+      logger.info(`Call off ordering party found in ORDAPI for ${orderId}: ${callOffOrgData}`);
+      return getContext({ orderId, data: callOffOrgData });
+    }
+    logger.error(`No call off ordering party data returned from ORDAPI for ${orderId}`);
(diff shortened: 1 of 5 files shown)


app/pages/sections/call-off-ordering-party/controller.js
```diff
@@ -3,14 +3,29 @@ import { getContext } from './contextCreator';
...
+    const callOffOrgData = await getData({ endpoint: callOffOrgDataEndpoint, accessToken, logger });
 
-  if (callOffOrgData) {
-    logger.info(`Call off ordering party found in ORDAPI for ${orderId}: ${callOffOrgData}`);
-    return getContext({ orderId, data: callOffOrgData });
+    if (callOffOrgData) {
+      logger.info(`Call off ordering party found in ORDAPI for ${orderId}: ${callOffOrgData}`);
+      return getContext({ orderId, data: callOffOrgData });
+    }
+    logger.error(`No call off ordering party data returned from ORDAPI for ${orderId}`);
```
alibombaye, line 15: should we be logging this as `error` it might be perfectly fine behaviour for it not to be there

app/pages/sections/call-off-ordering-party/controller.js
```diff
@@ -3,14 +3,29 @@ import { getContext } from './contextCreator';
...
+    throw new Error();
```
alibombaye, line 16: why throw and then catch the Error we can just proceed with the rest of the logic as is?

app/pages/sections/call-off-ordering-party/controller.js
```diff
@@ -3,14 +3,29 @@ import { getContext } from './contextCreator';
...
+  } catch (err) {
+    logger.info(`No call off ordering party found in ORDAPI for ${orderId}. ${err}`);
+
+    try {
+      const orgDataEndpoint = getEndpoint({ endpointLocator: 'getOrganisationById', options: { orgId } });
+      const orgData = await getData({ endpoint: orgDataEndpoint, accessToken, logger });
+      // const orgData = mockOrgData;
```
alibombaye: comment

app/pages/sections/call-off-ordering-party/controller.js
kat1906: I'll remove it as I don't think it should ever be reachable anyway. If not found the api will send error 404

app/pages/sections/call-off-ordering-party/controller.test.js
```diff
@@ -26,36 +26,87 @@ const mockOrderingPartyData = {
...
+        expect(getData).toHaveBeenNthCalledWith(1, {
+          endpoint: `${orderApiUrl}/api/v1/orders/order-id/sections/ordering-party`,
+          accessToken: 'access_token',
+          logger,
+        });
+        expect(getData).toHaveBeenNthCalledWith(2, {
+          endpoint: `${organisationApiUrl}/api/v1/Organisations/org-id`,
+          accessToken: 'access_token',
+          logger,
+        });
```
alibombaye, line 70: 👍 I like this

app/pages/sections/call-off-ordering-party/controller.js
kat1906: see above

app/pages/sections/call-off-ordering-party/controller.js
kat1906: I already removed it

app/pages/sections/call-off-ordering-party/controller.test.js
```diff
@@ -26,36 +26,87 @@ const mockOrderingPartyData = {
...
       });
     });
 
-    it('should call getContext with the correct params when data returned from getData', async () => {
-      getData
-        .mockResolvedValueOnce(mockOrderingPartyData);
-      contextCreator.getContext
-        .mockResolvedValueOnce();
+    describe('when call-off-ordering-party is already completed', () => {
+      it('should call getData twice with the correct params', async () => {
```
alibombaye, line 88: should be called just once right?

kat1906: ah yeah didnt rename the test

app/pages/sections/call-off-ordering-party/controller.test.js
```diff
@@ -26,36 +26,87 @@ const mockOrderingPartyData = {
...
+        expect(getData.mock.calls.length).toEqual(1);
+        expect(getData).toHaveBeenCalledWith({
+          endpoint: `${orderApiUrl}/api/v1/orders/order-id/sections/ordering-party`,
+          accessToken: 'access_token',
+          logger,
+        });
+      });
 
-      await getCallOffOrderingPartyContext({ orderId: 'order-id', accessToken: 'access_token' });
+      it('should call getContext with the correct params when data returned from organisations API', async () => {
```
alibombaye, line 101: when returned from the orderingApi?

app/pages/sections/call-off-ordering-party/controller.js
alibombaye: I don't think we need to throw a new error then. The 404 will throw the error and the catch then should work.

kat1906: yep removed

kat1906: had to put throw at end to throw the inner errors out

PR #66, opened 2020-05-20
4615: Populate ordering party data

kat1906:
Story: [AB#4615](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/4615)
Task: [AB#7087](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/7087)


kat1906:
even better - the null line just doesn't show!

PR #70, opened 2020-05-21
Feature/4840 call bapi

alibombaye:
Story [AB#4840](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/4840)
Task [AB#7099](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/7099)

This PR is to call bapi to get the list of suppliers.

Subsequent PRs will be showing the relevant page based upon the whether suppliers were found or not

### DIFF
--- a/app/config.js
+++ b/app/config.js
@@ -23,6 +23,9 @@ module.exports = {
   // Organisations API
   organisationApiUrl: process.env.ORGANISATION_API_URL || 'http://localhost:5103',
 
+  // Buying Catalogue API
+  buyingCatalogueApiHost: process.env.API_HOST || 'http://localhost:5100',
+
   // BLOBSTORE_HOST
   blobstoreHost: process.env.BLOBSTORE_HOST || 'https://gpitfuturesdevsa.blob.core.windows.net',
 
--- a/app/routes.js
+++ b/app/routes.js
@@ -7,7 +7,12 @@ import { logger } from './logger';
 import { withCatch, getHealthCheckDependencies, extractAccessToken } from './helpers/routerHelper';
 import { getDashboardContext } from './pages/dashboard/controller';
 import { getDescriptionContext, getDescriptionErrorContext, postOrPutDescription } from './pages/sections/description/controller';
-import * as supplierSearchController from './pages/sections/supplier/search/controller';
+import {
+  getSupplierSearchPageContext,
+  validateSupplierSearchForm,
+  findSuppliers,
+  getSupplierSearchPageErrorContext,
+} from './pages/sections/supplier/search/controller';
 import includesContext from './includes/manifest.json';
 import { getTaskListPageContext } from './pages/task-list/controller';
 import { getCallOffOrderingPartyContext } from './pages/sections/call-off-ordering-party/controller';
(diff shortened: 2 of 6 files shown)


app/config.js
```diff
@@ -23,6 +23,9 @@ module.exports = {
   // Organisations API
   organisationApiUrl: process.env.ORGANISATION_API_URL || 'http://localhost:5103',
 
+  // Buying Catalogue API
+  buyingCatalogueApiHost: process.env.API_HOST || 'http://localhost:5100',
```
kat1906, line 27: we can both get demo on adding these today as I have to add orgapi too!

alibombaye: that would be good. We've got a task to add them to the cluster. See if Adrian has any time to go through that today.

app/routes.js
```diff
@@ -7,7 +7,7 @@ import { logger } from './logger';
 import { withCatch, getHealthCheckDependencies, extractAccessToken } from './helpers/routerHelper';
 import { getDashboardContext } from './pages/dashboard/controller';
 import { getDescriptionContext, getDescriptionErrorContext, postOrPutDescription } from './pages/sections/description/controller';
-import { getSupplierSearchPageContext, validateSupplierSearchForm, getSupplierSearchPageErrorContext } from './pages/sections/supplier/search/controller';
+import * as supplierSearchController from './pages/sections/supplier/search/controller';
```
kat1906, line 10: how come you are importing like this?

alibombaye: it wanted them all on separate lines since it was too long

alibombaye: can change it just adds lot of lines of imports at the top and this routes file is just going it get even bigger.

app/routes.js
```diff
@@ -88,20 +88,29 @@ export const routes = (authProvider) => {
...
   router.post('/organisation/:orderId/supplier/search', authProvider.authorise({ claim: 'ordering' }), withCatch(authProvider, async (req, res) => {
     const { orderId } = req.params;
 
-    const response = validateSupplierSearchForm({ data: req.body });
+    const response = supplierSearchController.validateSupplierSearchForm({ data: req.body });
 
-    const context = await getSupplierSearchPageErrorContext({
-      orderId,
-      validationErrors: response.errors,
+    if (!response.success) {
```
kat1906, line 100: I think the pattern we usually use is happy first?

alibombaye: yes I noticed this but in this instance this is the validation we are doing. So we don't want to progress any further in the logic of this route if there are validation errors. 

So that's why I've had to change the pattern here slightly.

alibombaye: actually going to try this the other way and see how that looks

alibombaye: yeah looks better 👍

app/routes.js
kat1906: I think that is preferable to the massively long names when you call the functions??

kat1906: especially as it will probably push some of the lines over the char limit then will have to have multiple lines there instead

PR #73, opened 2020-05-21
4616: Save data

kat1906:
Story: [AB#4616](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/4616)
Task: [AB#7090](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/7090)

### DIFF
--- a/app/pages/sections/call-off-ordering-party/controller.js
+++ b/app/pages/sections/call-off-ordering-party/controller.js
@@ -1,7 +1,31 @@
-import { getData } from 'buying-catalogue-library';
+import { getData, putData } from 'buying-catalogue-library';
 import { getContext } from './contextCreator';
 import { getEndpoint } from '../../../endpoints';
 import { logger } from '../../../logger';
+
+const formatPutData = data => ({
+  organisation: {
+    name: data.name,
+    odsCode: data.odsCode,
+    address: {
+      line1: data.line1,
+      line2: data.line2,
+      line3: data.line3,
+      line4: data.line4,
+      line5: data.line5,
+      town: data.town,
+      county: data.county,
+      postcode: data.postcode,
+      country: data.country,
+    },
+  },
+  primaryContact: {
+    firstName: data.firstName,
+    lastName: data.lastName,
+    emailAddress: data.emailAddress,
+    telephoneNumber: data.telephoneNumber,
+  },
+});
 
 export const getCallOffOrderingPartyContext = async ({ orderId, orgId, accessToken }) => {
   try {
@@ -26,4 +50,26 @@ export const getCallOffOrderingPartyContext = async ({ orderId, orgId, accessTok
     }
   }
   throw new Error();
+};
+
+export const putCallOffOrderingParty = async ({
+  orgId, orderId, data, accessToken,
+}) => {
+  const endpoint = getEndpoint({ endpointLocator: 'putOrderingParty', options: { orderId } });
+  const body = formatPutData(data);
+  try {
+    await putData({
+      endpoint,
+      body,
+      organisationId: orgId,
+      accessToken,
+      logger,
+    });
+
+    logger.info(`Call off ordering party updated - order id: ${orderId}, ${JSON.stringify(data)}`);
+    return { success: true };
+  } catch (err) {
+    logger.error('Error updating call-off-ordering-party for order');
+    throw new Error();
+  }
 };
--- a/app/pages/sections/call-off-ordering-party/controller.js
+++ b/app/pages/sections/call-off-ordering-party/controller.js
@@ -5,25 +5,25 @@ import { logger } from '../../../logger';
 
 const formatPutData = data => ({
   organisation: {
-    name: data.name,
-    odsCode: data.odsCode,
+    name: data.name ? data.name.trim() : null,
+    odsCode: data.odsCode ? data.odsCode.trim() : null,
     address: {
-      line1: data.line1,
-      line2: data.line2,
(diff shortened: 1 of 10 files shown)


app/pages/sections/call-off-ordering-party/controller.js
```diff
@@ -1,8 +1,32 @@
...
+      country: data.country ? data.country.trim() : null,
+    },
+  },
+  primaryContact: {
+    firstName: data.firstName ? data.firstName.trim() : null,
+    lastName: data.lastName ? data.lastName.trim() : null,
+    emailAddress: data.emailAddress ? data.emailAddress.trim() : null,
+    telephoneNumber: data.telephoneNumber ? data.telephoneNumber.trim() : null,
+  },
+});
```
alibombaye, line 28: In the template for this page. If you set the id of these fields and prefix it to include its parent id eg `address` or `primaryContact` then this should give you this structure when you come to submit it. Rather than a flat structure that you are getting at the moment and then having to do this format.

app/pages/sections/call-off-ordering-party/template.test.js
```diff
@@ -116,6 +116,55 @@ describe('call-off-ordering-party page', () => {
...
+      expect(townFormElement.attr('value')).toEqual(context.address.town);
+      expect(countyFormElement.length).toEqual(1);
+      expect(countyFormElement.attr('type')).toEqual('hidden');
+      expect(countyFormElement.attr('value')).toEqual(context.address.county);
+      expect(postcodeFormElement.length).toEqual(1);
+      expect(postcodeFormElement.attr('type')).toEqual('hidden');
+      expect(postcodeFormElement.attr('value')).toEqual(context.address.postcode);
+      expect(countryFormElement.length).toEqual(1);
+      expect(countryFormElement.attr('type')).toEqual('hidden');
+      expect(countryFormElement.attr('value')).toEqual(context.address.country);
```
alibombaye, line 164: Can you put a blank line in between each element expectation just so they are grouped and easier to read.

app/routes.test.js
```diff
@@ -300,6 +303,89 @@ describe('routes', () => {
...
+      const { cookies, csrfToken } = await getCsrfTokenFromGet({
+        app: request(setUpFakeApp()), csrfPagePath: path, mockAuthorisedCookie,
+      });
+
+      return request(setUpFakeApp())
+        .post(path)
+        .type('form')
+        .set('Cookie', [cookies, mockAuthorisedCookie])
+        .send({
+          description: 'a description of the order',
```
alibombaye, line 354: don't think these affects the test but suppose this should be call-off ordering party data instead. Or just remove it all together so it doesn't get confusing

app/pages/sections/call-off-ordering-party/controller.test.js
```diff
@@ -45,6 +45,47 @@ const mockOrgData = {
   },
   catalogueAgreementSigned: false,
 };
+const mockPutData = {
```
alibombaye, line 48: think this should be renamed to `mockFormData` perhaps. Since this is what we get from the form when it's submitted.

app/routes.test.js
```diff
@@ -300,6 +303,89 @@ describe('routes', () => {
...
+      const { cookies, csrfToken } = await getCsrfTokenFromGet({
+        app: request(setUpFakeApp()), csrfPagePath: path, mockAuthorisedCookie,
+      });
+
+      return request(setUpFakeApp())
+        .post(path)
+        .type('form')
+        .set('Cookie', [cookies, mockAuthorisedCookie])
+        .send({
+          description: 'a lovely decription',
```
alibombaye, line 378: I think we should remove these from the routes test since it doesn't have any affect to the test.

app/pages/sections/call-off-ordering-party/controller.js
kat1906: ooh interesting

app/pages/sections/call-off-ordering-party/template.test.js
kat1906: sure

app/pages/sections/call-off-ordering-party/controller.js
alibombaye: On second thoughts I'm not sure it's going to be actually worth it. I'd leave as is for now. It's better I think that we have control on the structure of the body as you have done. Rather than rely on the form.

app/routes.test.js
kat1906: ah yeah

app/routes.test.js
kat1906: ok

app/pages/sections/call-off-ordering-party/controller.test.js
kat1906: sure 🤷

app/pages/sections/call-off-ordering-party/controller.js
kat1906: ok that is easier for me 😄

alibombaye:
The pipeline build issue should get sorted when you merge latest development

app/pages/sections/call-off-ordering-party/template.test.js
kat1906: I was going to do it all fancy but ... cba 😄

PR #75, opened 2020-05-22
Feature/4840 show suppliers returned

alibombaye:
Story [AB#4840](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/4840)
Task [AB#7102](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/7102)

Just a note I had to move the actual finding of suppliers to the /select route. Since we won't be able to share the suppliers found from /search to /select without doing a search again in /select.

Now the /search just validates that something is entered and if so redirects to /select with the query string appended of what was entered.

Screenshot 2020-05-22 at 10 24 20

### DIFF
--- a/app/pages/sections/supplier/select/contextCreator.js
+++ b/app/pages/sections/supplier/select/contextCreator.js
@@ -1,9 +1,28 @@
 import manifest from './manifest.json';
 import { baseUrl } from '../../../../config';
 
-export const getContext = ({ orderId }) => {
+const generateSupplierOptions = ({ suppliers }) => {
+  const supplierOptions = suppliers.map(supplier => ({
+    value: supplier.supplierId,
+    text: supplier.name,
+  }));
+
+  return supplierOptions;
+};
+
+const generateQuestionsContext = ({ suppliers }) => {
+  const questions = manifest.questions.map(question => ({
+    ...question,
+    options: generateSupplierOptions({ suppliers }),
+  }));
+
+  return questions;
+};
+
+export const getContext = ({ orderId, suppliers }) => {
   const context = ({
     ...manifest,
+    questions: suppliers && generateQuestionsContext({ suppliers }),
     backLinkHref: `${baseUrl}/organisation/${orderId}/supplier/search`,
   });
   return context;
--- a/app/pages/sections/supplier/select/contextCreator.js
+++ b/app/pages/sections/supplier/select/contextCreator.js
@@ -1,29 +1,22 @@
 import manifest from './manifest.json';
 import { baseUrl } from '../../../../config';
 
-const generateSupplierOptions = ({ suppliers }) => {
-  const supplierOptions = suppliers.map(supplier => ({
+const generateSupplierOptions = ({ suppliers }) => (
+  suppliers.map(supplier => ({
     value: supplier.supplierId,
     text: supplier.name,
-  }));
+  }))
+);
 
-  return supplierOptions;
-};
-
-const generateQuestionsContext = ({ suppliers }) => {
-  const questions = manifest.questions.map(question => ({
+const generateQuestionsContext = ({ suppliers }) => (
+  manifest.questions.map(question => ({
     ...question,
     options: generateSupplierOptions({ suppliers }),
-  }));
+  }))
+);
 
-  return questions;
-};
-
-export const getContext = ({ orderId, suppliers }) => {
-  const context = ({
-    ...manifest,
-    questions: suppliers && generateQuestionsContext({ suppliers }),
-    backLinkHref: `${baseUrl}/organisation/${orderId}/supplier/search`,
-  });
-  return context;
-};
+export const getContext = ({ orderId, suppliers }) => ({
+  ...manifest,
+  questions: suppliers && generateQuestionsContext({ suppliers }),
+  backLinkHref: `${baseUrl}/organisation/${orderId}/supplier/search`,
+});
--- a/app/pages/sections/supplier/select/contextCreator.js
+++ b/app/pages/sections/supplier/select/contextCreator.js
@@ -1,0 +1,10 @@
+import manifest from './manifest.json';
+import { baseUrl } from '../../../../config';
+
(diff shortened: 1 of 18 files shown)


alibombaye:
We may want to change the route of the /select to /search/:suppliersearchstring
rather than pass if through via a query string.

I am thinking of when we want to go back from the supplier section page to the select page. We will need this suppliersearchstring again to do the find.

Or do we think now is a good time to start thinking about using redis state?

app/pages/sections/supplier/select/contextCreator.js
```diff
@@ -0,0 +1,29 @@
+import manifest from './manifest.json';
+import { baseUrl } from '../../../../config';
+
+const generateSupplierOptions = ({ suppliers }) => {
+  const supplierOptions = suppliers.map(supplier => ({
```
kat1906, line 5: can just return from this line:
const generateSupplierOptions = ({ suppliers }) => suppliers.map(supplier => ({
    value: supplier.supplierId,
    text: supplier.name,
  }));

app/pages/sections/supplier/select/contextCreator.js
```diff
@@ -0,0 +1,29 @@
...
+    value: supplier.supplierId,
+    text: supplier.name,
+  }));
+
+  return supplierOptions;
+};
+
+const generateQuestionsContext = ({ suppliers }) => {
+  const questions = manifest.questions.map(question => ({
```
kat1906, line 14: same here

app/pages/sections/supplier/select/contextCreator.js
```diff
@@ -0,0 +1,29 @@
...
+    ...question,
+    options: generateSupplierOptions({ suppliers }),
+  }));
+
+  return questions;
+};
+
+export const getContext = ({ orderId, suppliers }) => {
+  const context = ({
```
kat1906, line 23: and here

app/pages/sections/supplier/select/controller.js
```diff
@@ -0,0 +1,16 @@
...
+
+export const findSuppliers = async ({ supplierNameToFind, accessToken }) => {
+  const endpoint = getEndpoint({ endpointLocator: 'getSearchSuppliers', options: { supplierNameToFind } });
+  const suppliersFound = await getData({ endpoint, accessToken, logger });
+  logger.info(`Searching for "${supplierNameToFind}" returned ${suppliersFound.length} supplier`);
+
+  return suppliersFound;
+};
+
+export const getSupplierSelectPageContext = ({ orderId, suppliers }) => (
```
kat1906, line 14: can shorten this to:
export const getSupplierSelectPageContext = (params) => getContext(params)

app/pages/sections/supplier/select/ui.test.js
```diff
@@ -0,0 +1,177 @@
...
+    .expect(getLocation()).eql('http://identity-server/login');
+});
+
+test('should render Supplier select page', async (t) => {
+  nock(solutionsApiUrl)
+    .get('/api/v1/suppliers?name=some-supp')
+    .reply(200, [{}]);
+
+  await pageSetup(t, true);
+  await t.navigateTo(`${pageUrl}?supplierNameToFind=some-supp`);
```
kat1906, line 55: really don't like the query - just use name? it is not like we are using any others here

kat1906: this part is user facing after all

app/pages/sections/supplier/select/ui.test.js
```diff
@@ -0,0 +1,177 @@
...
+    .expect(page.exists).ok();
+});
+
+test('should navigate to /organisation/order-1/supplier/search when click on backlink', async (t) => {
+  nock(solutionsApiUrl)
+    .get('/api/v1/suppliers?name=some-supp')
+    .reply(200, [{}]);
+
+  await pageSetup(t, true);
+  await t.navigateTo(`${pageUrl}?supplierNameToFind=some-supp`);
```
kat1906, line 68: you have the pageUrl variable at the top, you could add another with the query string?

app/routes.js
```diff
@@ -137,12 +152,7 @@ export const routes = (authProvider) => {
       });
     }
 
-    const context = await getSupplierSearchPageErrorContext({
-      orderId,
-      validationErrors: response.errors,
-    });
-
-    return res.render('pages/sections/supplier/search/template.njk', addContext({ context, user: req.user, csrfToken: req.csrfToken() }));
+    return res.redirect(`${config.baseUrl}/organisation/${orderId}/supplier/search`);
```
kat1906, line 155: where is the redirect to the 0 found page?

app/pages/sections/supplier/select/ui.test.js
alibombaye: This will be removed once we have state. OK to leave as is for now?

app/routes.js
alibombaye: the throw new ErrorContext on like 146

app/pages/sections/supplier/select/ui.test.js
kat1906: I would like it changed (just in case) - you gan do global search and replace - will take 30secs

app/pages/sections/supplier/select/ui.test.js
alibombaye: again will be removed once we have state. OK to leave for now?

app/pages/sections/supplier/select/ui.test.js
alibombaye: haha ok fair enough. Like the vote of confidence 😢

app/pages/sections/supplier/select/ui.test.js
kat1906: global search and replace 😆

PR #82, opened 2020-05-26
4619: Validate date

kat1906:
Story: [AB#4619](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/4619)
Task: [AB#7125](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/7125)

JS `new Date` validates 31st Feb 2020 to 2nd March 2020 so need to check the month and year are correct.

`CommencementDateInvalid` - unsure of how this error is going to be used - after speaking with Luke it is for when characters apart from numbers are entered but the date input only allows numbers so may not be necessary.


app/pages/sections/commencement-date/controller.js
```diff
@@ -1,3 +1,10 @@
 import { getContext } from './contextCreator';
+import { getDateErrors } from './getDateErrors';
 
 export const getCommencementDateContext = async parmas => getContext(parmas);
+
+export const putCommencementDate = async ({ data }) => {
+  const errors = [getDateErrors(data)];
+  if (errors[0]) return { success: false, errors };
```
alibombaye, line 8: `errors.length > 0` perhaps?

kat1906: nope - counts the null as an entry in the array

app/pages/sections/commencement-date/getDateErrors.js
```diff
@@ -0,0 +1,44 @@
...
+      field: 'CommencementDate',
+      part: ['year'],
+      id: 'CommencementDateYearRequired',
+    },
+    {
+      field: 'CommencementDate',
+      part: ['day', 'month'],
+      id: 'CommencementDateNotReal',
+    },
+  ];
```
alibombaye, line 28: rather that this being an array. how about making this an object. Using the id as the key also.  That way when we do the lookup below we can use the id rather than the index value.

app/pages/sections/commencement-date/getDateErrors.test.js
```diff
@@ -0,0 +1,94 @@
...
+  it('should return CommencementDateNotReal error for "day" and "month" when day and month not numbers', () => {
+    const validationResponse = getDateErrors({ 'commencementDate-day': 'a', 'commencementDate-month': 'a', 'commencementDate-year': 'a' });
+    expect(validationResponse.field).toEqual('CommencementDate');
+    expect(validationResponse.id).toEqual('CommencementDateNotReal');
+    expect(validationResponse.part.length).toEqual(2);
+    expect(validationResponse.part.includes('day')).toBe(true);
+    expect(validationResponse.part.includes('month')).toBe(true);
+    expect(validationResponse.part.includes('year')).toBe(false);
+  });
+});
```
alibombaye, line 94: These are nice!!

app/pages/sections/commencement-date/getDateErrors.js
kat1906: okay dokey

kat1906: but the reason I did it like this is that it preserves the format that is needed for the errors to be processed later
so you want e.g. {
    CommencementDateDayRequired: {
      field: 'CommencementDate',
      part: ['day'],
      id: 'CommencementDateDayRequired',
    },
}

app/pages/sections/routes.js
```diff
@@ -74,5 +74,20 @@ export const sectionRoutes = (authProvider, addContext, sessionManager) => {
...
+    const { orderId } = req.params;
+
+    const response = await putCommencementDate({
+      orgId: req.user.primaryOrganisationId,
+      orderId,
+      data: req.body,
+      accessToken: extractAccessToken({ req, tokenType: 'access' }),
+    });
+
+    if (!response.success) logger.info(`validation error: ${JSON.stringify(response)}`);
```
alibombaye, line 87: take it you are going to change this later on to follow the pattern of handling the happy path?

kat1906: yep

app/pages/sections/routes.test.js
```diff
@@ -311,4 +315,82 @@ describe('routes', () => {
...
+      return request(setUpFakeApp())
+        .post(path)
+        .type('form')
+        .set('Cookie', [cookies, mockAuthorisedCookie])
+        .send({ _csrf: csrfToken })
+        .expect(302)
+        .then((res) => {
+          expect(res.redirect).toEqual(true);
+          expect(res.headers.location).toEqual(`${baseUrl}/organisation/order-id`);
+          expect(res.text.includes('data-test-id="error-title"')).toEqual(false);
```
alibombaye, line 392: Just so you know that this test is not right for when you do have the validation fully plugged in.

kat1906: yeah I know but it is right for now

app/pages/sections/commencement-date/getDateErrors.js
alibombaye: yes something like that is perfect. I don't mind the id being duplicated but just to make the look up look cleaner and not use index values

PR #107, opened 2020-06-01
rename call-off-ordering-party to ordering-party

alibombaye:
Story [AB#4616](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/4616)
Task [AB#7093](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/7093)

update id of section to match what is returned by API so the section can be marked as complete


kat1906:
I did think about doing this 😄

PR #109, opened 2020-06-01
format date from commencement date

alibombaye:
Story [AB#4619](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/4619)
Task [AB#7130](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/7130)

The commencementDate from the API returns it like this "commencementDate": "2020-06-18T00:00:00"

We are not expecting the Timestamp bit which is why the Day was not showing. This fix will format the date returned from the API to one we can work with

### DIFF
--- a/app/helpers/dateFormatter.js
+++ b/app/helpers/dateFormatter.js
@@ -6,3 +6,12 @@ export const formatDate = (date) => {
   const da = new Intl.DateTimeFormat('en', { day: 'numeric' }).format(d);
   return `${da} ${mo} ${ye}`;
 };
+
+export const formatCommencementDate = (date) => {
+  const d = new Date(date);
+  if (Number.isNaN(d.getTime())) return '';
+  const ye = new Intl.DateTimeFormat('en', { year: 'numeric' }).format(d);
+  const mo = new Intl.DateTimeFormat('en', { month: '2-digit' }).format(d);
+  const da = new Intl.DateTimeFormat('en', { day: '2-digit' }).format(d);
+  return `${da} ${mo} ${ye}`;
+};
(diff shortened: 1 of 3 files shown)


app/pages/sections/commencement-date/contextCreator.js
```diff
@@ -1,9 +1,10 @@
 import manifest from './manifest.json';
 import { baseUrl } from '../../../config';
+import { formatCommencementDate } from '../../../helpers/dateFormatter';
 
 const addDataToManifest = (commencementDate) => {
   if (commencementDate) {
-    const [year, month, day] = commencementDate.split('-');
+    const [day, month, year] = formatCommencementDate(commencementDate).split(' ');
```
kat1906, line 7: might as well return the array from this function `const [day, month, year] = formatCommencementDate(commencementDate)`

app/helpers/dateFormatter.js
```diff
@@ -6,3 +6,12 @@ export const formatDate = (date) => {
...
   return `${da} ${mo} ${ye}`;
 };
+
+export const formatCommencementDate = (date) => {
+  const d = new Date(date);
+  if (Number.isNaN(d.getTime())) return '';
+  const ye = new Intl.DateTimeFormat('en', { year: 'numeric' }).format(d);
+  const mo = new Intl.DateTimeFormat('en', { month: '2-digit' }).format(d);
+  const da = new Intl.DateTimeFormat('en', { day: '2-digit' }).format(d);
+  return `${da} ${mo} ${ye}`;
```
kat1906, line 16: `return [da, mo, ye]` here

PR #125, opened 2020-06-04
7412 display service table

samMcCarthyBJSS:
Story [AB#7412](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/7412)
Task [AB#7508](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/7508)

### DIFF
--- a/app/pages/sections/service-recipients/contextCreator.js
+++ b/app/pages/sections/service-recipients/contextCreator.js
@@ -1,8 +1,21 @@
 import manifest from './manifest.json';
 import { baseUrl } from '../../../config';
 
-export const getContext = ({ orderId }) => ({
-  ...manifest,
-  title: `${manifest.title} ${orderId}`,
-  backLinkHref: `${baseUrl}/organisation/${orderId}`,
-});
+export const getContext = ({ orderId, serviceRecipientsData = [] }) => {
+  const tableData = serviceRecipientsData.map(data => ({
+    organisationName: {
+      id: `${data.odsCode}-id`,
+      name: `${data.odsCode}-name`,
+      value: data.odsCode,
+      text: data.name,
+      checked: false,
+    },
+    odsCode: data.odsCode,
+  }));
+  return {
+    ...manifest,
+    title: `${manifest.title} ${orderId}`,
+    backLinkHref: `${baseUrl}/organisation/${orderId}`,
+    tableData,
+  };
+};
--- a/app/pages/sections/service-recipients/template.njk
+++ b/app/pages/sections/service-recipients/template.njk
@@ -41,7 +41,6 @@
     </div>
 
     <hr class="bc-c-order-table__hr nhsuk-grid-column-full nhsuk-u-margin-top-4 nhsuk-u-margin-bottom-4"/>
-    
 
     <div class="nhsuk-grid-row">
       <div class="nhsuk-grid-column-two-thirds nhsuk-u-font-size-16">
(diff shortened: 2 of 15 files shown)


app/pages/sections/service-recipients/contextCreator.js
```diff
@@ -1,8 +1,21 @@
 import manifest from './manifest.json';
 import { baseUrl } from '../../../config';
 
-export const getContext = ({ orderId }) => ({
-  ...manifest,
-  title: `${manifest.title} ${orderId}`,
-  backLinkHref: `${baseUrl}/organisation/${orderId}`,
-});
+export const getContext = ({ orderId, serviceRecipientsData = [] }) => {
+  const tableData = serviceRecipientsData.map(data => ({
```
alibombaye, line 5: would be quite nice to extract this to a little helper function.

samMcCarthyBJSS: we'll do it in the next pr as that's modifying the checkbox return

app/pages/sections/service-recipients/template.njk
```diff
@@ -29,20 +30,32 @@
...
+            <div data-test-id="organisation-name-checkbox-{{row.odsCode}}" class="nhsuk-grid-column-two-thirds">
+              {{ checkboxes({
+                  items: [row.organisationName]
+                }) }}
+            </div>
+            <div data-test-id="ods-code-{{row.odsCode}}" class="nhsuk-grid-column-one-third bc-c-service-recipients-table__ods">
+              {{ row.odsCode }}
+            </div>
+          </div>
+        {% endfor %}
```
alibombaye, line 58: I thought we were going to use the table component we have here?

samMcCarthyBJSS: its because it line rows in between every row
and no other component would have a table with no lines

app/pages/sections/service-recipients/ui-tests/general.ui.test.js
```diff
@@ -0,0 +1,191 @@
...
+  },
+  {
+    name: 'Some service recipient 2',
+    odsCode: 'ods2',
+  },
+];
+
+const getLocation = ClientFunction(() => document.location.href);
+
+fixture('service-recipients page')
```
alibombaye, line 46: rename this to `service-recipients page - general`

app/pages/sections/service-recipients/template.njk
alibombaye: Sorry I don't quite understand that. I thought we had proven out that our table component would be able to handle this sort of data coming in

kat1906: we checked with Lorraine and it is the only table that looks like this so rather than messing around with the component we just went with the easiest. option of rendering it in 2 columns like this

kat1906: it will accept the checkboxes fine but puts lines between each one

alibombaye: ok couldn't we just update the component to pass a flag so them lines are not shown. We've gone through the trouble of creating the component. We may as well use it.

kat1906: we are going to use it latee with the input fields - it is more hassle than it is worth to put that flag in 😄

kat1906: tech debt?

PR #142, opened 2020-06-09
basic select recipient page

alibombaye:
Story [AB#7261](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/7261)
Task [AB#7644](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/7644)

Screenshot 2020-06-09 at 17 21 50

### DIFF
--- a/app/pages/sections/catalogue-solutions/routes.js
+++ b/app/pages/sections/catalogue-solutions/routes.js
@@ -16,6 +16,9 @@ import {
 import {
   getSolutionPricePageContext,
 } from './select-price/controller';
+import {
+  getSolutionRecipientPageContext,
+} from './select-recipient/controller';
 
 const router = express.Router({ mergeParams: true });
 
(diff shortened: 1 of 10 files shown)


app/pages/sections/catalogue-solutions/routes.js
```diff
@@ -91,8 +94,10 @@ export const catalogueSolutionsRoutes = (authProvider, addContext, sessionManage
   router.get('/select-solution/select-price/select-recipient', authProvider.authorise({ claim: 'ordering' }), withCatch(authProvider, async (req, res) => {
     const { orderId } = req.params;
 
+    const context = await getSolutionRecipientPageContext({ orderId, solutionName: 'Solution One' });
```
kat1906, line 97: where should this solution name be coming from?

alibombaye: I'm going to do this part of the next PR but it will be coming from session. When a solution is selected we'll also need to store the solutionName too and retrieve that here 😢

app/pages/sections/catalogue-solutions/select-recipient/contextCreator.test.js
```diff
@@ -0,0 +1,38 @@
+import manifest from './manifest.json';
+import { getContext } from './contextCreator';
+import { baseUrl } from '../../../../config';
+
+jest.mock('../../getSectionErrorContext', () => ({
```
kat1906, line 5: don't think this needs mocked

alibombaye: ooops already merged.. I'll remove for the next PR and then re-add it back in for the PR after that 😄

PR #150, opened 2020-06-11
move solution, recipient and price to select folder

alibombaye:
Screenshot 2020-06-11 at 12 21 26

Moves solution, price and recipient to select sub folder and also the routes for this journey.

Also the urls are now 
/catalogue-solutions/select/solution
/catalogue-solutions/select/solution/price
/catalogue-solutions/select/solution/price/recipient


kat1906:
I love what you have done with the urls here 😄

PR #173, opened 2020-06-25
Feature/5127 show empty tables

alibombaye:
Story [AB#5127](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/5127)
Task [AB#7874](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/7874)

Screenshot 2020-06-25 at 14 09 21

### DIFF
--- a/app/pages/preview/manifest.json
+++ b/app/pages/preview/manifest.json
@@ -49,5 +49,37 @@
     ]
   },
   "recurringCostHeading": "Order items (recurring cost)",
-  "recurringCostDescription": "The following items carry a recurring annual cost for the duration of this Call-off Agreement. They're the Catalogue Solutions, Additional Services and Associated Services in this order that you'll be charged for at a monthly rate."
+  "recurringCostDescription": "The following items carry a recurring annual cost for the duration of this Call-off Agreement. They're the Catalogue Solutions, Additional Services and Associated Services in this order that you'll be charged for at a monthly rate.",
+  "recurringCostTable": {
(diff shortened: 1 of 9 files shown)


app/pages/preview/manifest.json
```diff
@@ -7,13 +7,92 @@
...
-  "commencementDateLabel": "Commencement date of the Call-off Agreement:"
+  "commencementDateLabel": "Commencement date of the Call-off Agreement:",
+  "oneOffCostHeading": "Order items (one-off cost)",
+  "oneOffCostDescription": "The following items carry a one-off cost for the duration of this Call-off Agreement. They're the Associated Services in this order you can use as and when required",
+  "oneOffCostTable": {
+    "columnInfo": [
+      {
+        "data": "Recipient name (ODS code)",
+        "width": "20%",
+        "classes": "nhsuk-u-font-size-14"
```
samMcCarthyBJSS, line 27: could the column class be used instead of having the fonts duplicated

alibombaye: that would be nice just to set this size once. But unfortunately that doesn't work. This class needs to be set on each on each cell. I think perhaps the default nhs styling on these cells then override the one set at the top level 🤷

samMcCarthyBJSS: ah right ok that's fair

PR #176, opened 2020-06-26
7840 fe validation

samMcCarthyBJSS:
Story [AB#7840](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/7840)
Task [AB#7722](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/7722)

image

### DIFF
--- a/app/helpers/questionExtractor.test.js
+++ b/app/helpers/questionExtractor.test.js
@@ -1,0 +1,19 @@
+import { questionExtractor } from './questionExtractor';
+
+const manifest = {
+  questions: [
+    {
+      id: 'test-id',
+      data: 'test data',
+    },
+  ],
+};
+describe('questionExtractor', () => {
+  it('returns the correct question', () => {
+    expect(questionExtractor('test-id', manifest)).toEqual(manifest.questions[0]);
+  });
+
+  it('does not return a question if there is no match', () => {
+    expect(questionExtractor('wrong-id', manifest)).toEqual(undefined);
+  });
+});
--- a/app/pages/sections/catalogue-solutions/order-item/contextCreator.js
+++ b/app/pages/sections/catalogue-solutions/order-item/contextCreator.js
@@ -18,6 +18,11 @@ export const populateTable = ((selectedPrice) => {
 
 export const formatFormData = ((populatedData) => {
   manifest.questions[2].data = populatedData.quantity ? populatedData.quantity.trim() : '';
+  if (populatedData.price) {
+    manifest.addPriceTable.data[0][0].question.data = populatedData.price.trim();
+  } else {
+    manifest.addPriceTable.data[0][0].question.data = '';
+  }
 });
 
 export const getContext = ({
--- a/app/pages/sections/catalogue-solutions/order-item/contextCreator.js
+++ b/app/pages/sections/catalogue-solutions/order-item/contextCreator.js
@@ -1,6 +1,5 @@
 import manifest from './manifest.json';
 import { baseUrl } from '../../../../config';
-import { logger } from '../../../../logger';
 
 export const addData = ((selectedPrice) => {
   manifest.questions.estimationPeriod.options.forEach((option, i) => {
--- a/app/pages/sections/catalogue-solutions/order-item/contextCreator.js
+++ b/app/pages/sections/catalogue-solutions/order-item/contextCreator.js
@@ -1,10 +1,11 @@
 import manifest from './manifest.json';
 import { baseUrl } from '../../../../config';
 import { getSectionErrorContext } from '../../getSectionErrorContext';
+import { questionExtractor } from '../../../../helpers/questionExtractor';
 
 export const populateEstimationPeriod = ((selectedPrice) => {
-  manifest.questions[2].options.forEach((option, i) => {
-    manifest.questions[2].options[i]
+  questionExtractor('selectEstimationPeriod', manifest).options.forEach((option, i) => {
+    questionExtractor('selectEstimationPeriod', manifest).options[i]
       .checked = option.text.toLowerCase() === selectedPrice
         .timeUnit.description.toLowerCase()
         ? true : undefined;
@@ -17,7 +18,7 @@ export const populateTable = ((selectedPrice) => {
 });
 
 export const formatFormData = ((populatedData) => {
-  manifest.questions[2].data = populatedData.quantity ? populatedData.quantity.trim() : '';
+  questionExtractor('quantity', manifest).data = populatedData.quantity ? populatedData.quantity.trim() : '';
   if (populatedData.price) {
     manifest.addPriceTable.data[0][0].question.data = populatedData.price.trim();
   } else {
--- a/app/pages/sections/catalogue-solutions/order-item/contextCreator.js
+++ b/app/pages/sections/catalogue-solutions/order-item/contextCreator.js
@@ -3,8 +3,8 @@ import { baseUrl } from '../../../../config';
 import { getSectionErrorContext } from '../../getSectionErrorContext';
 
 export const populateEstimationPeriod = ((selectedPrice) => {
-  manifest.questions.estimationPeriod.options.forEach((option, i) => {
-    manifest.questions.estimationPeriod.options[i]
+  manifest.questions[2].options.forEach((option, i) => {
+    manifest.questions[2].options[i]
       .checked = option.text.toLowerCase() === selectedPrice
         .timeUnit.description.toLowerCase()
         ? true : undefined;
@@ -17,7 +17,7 @@ export const populateTable = ((selectedPrice) => {
 });
 
 export const formatFormData = ((populatedData) => {
-  manifest.questions.quantity.data = populatedData.quantity ? populatedData.quantity.trim() : '';
+  manifest.questions[2].data = populatedData.quantity ? populatedData.quantity.trim() : '';
 });
 
 export const getContext = ({
@@ -25,7 +25,7 @@ export const getContext = ({
 }) => {
   populateEstimationPeriod(selectedPrice);
   populateTable(selectedPrice);
-  // if (populatedData) formatFormData(populatedData);
+  if (populatedData) formatFormData(populatedData);
 
   return ({
     ...manifest,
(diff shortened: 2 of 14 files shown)


app/pages/sections/catalogue-solutions/order-item/contextCreator.js
```diff
@@ -1,9 +1,10 @@
 import manifest from './manifest.json';
 import { baseUrl } from '../../../../config';
+import { getSectionErrorContext } from '../../getSectionErrorContext';
 
 export const populateEstimationPeriod = ((selectedPrice) => {
-  manifest.questions.estimationPeriod.options.forEach((option, i) => {
-    manifest.questions.estimationPeriod.options[i]
+  manifest.questions[2].options.forEach((option, i) => {
```
alibombaye, line 6: 😢 It's a shame that having our manifest with the question as a dictionary rather than an array is so much work. I don't like hardcoding these number. If in a few months a decision is made to add another question before this then this part will fall apart. Can you create a helper function within the helpers folder to find the question in the manifest passing in the questionId. Hopefully that way we can avoid using these hardcoded numbers.

app/pages/sections/catalogue-solutions/order-item/template.njk
```diff
@@ -19,34 +19,39 @@
...
-                question: questions.estimationPeriod
-              })
-            }}
-            {{ viewExpandableSection(questions.estimationPeriod.expandableSection) }}
+              {{
+                radiobuttonOptions({
+                  question: questions[2]
+                })
+              }}
+              {{ viewExpandableSection(questions[2].expandableSection) }}
```
alibombaye, line 54: 😞 preferred what you had before but we are just going to have to live with this for now

app/pages/sections/catalogue-solutions/order-item/template.test.js
```diff
@@ -23,6 +23,57 @@ describe('catalogue-solutions order-item page', () => {
     });
   }));
 
+  it('should render error summary with correct error text and hrefs if there are errors', componentTester(setup, (harness) => {
+    const context = {
+      questions: [
+        {},
```
alibombaye, line 29: what do these empty objects refer to?

alibombaye: OK I know now. These are the other questions we need to pass into the context. Cool this is fine 👍

app/pages/sections/catalogue-solutions/order-item/contextCreator.js
```diff
@@ -15,11 +17,21 @@ export const populateTable = ((selectedPrice) => {
...
+  if (populatedData.price) {
+    manifest.addPriceTable.data[0][0].question.data = populatedData.price.trim();
+  } else {
+    manifest.addPriceTable.data[0][0].question.data = '';
+  }
+});
+
 export const getContext = ({
-  orderId, solutionName, serviceRecipientName, odsCode, selectedPrice,
+  orderId, solutionName, serviceRecipientName, odsCode, selectedPrice, populatedData = null,
```
alibombaye, line 30: why is this set to a default value of null?

app/pages/sections/catalogue-solutions/order-item/contextCreator.test.js
```diff
@@ -47,43 +52,46 @@ describe('catalogue-solutions order-item contextCreator', () => {
...
-      };
+      },
+      {
+        type: 'input',
+        id: 'price',
+      },
+      ];
 
       const context = getContext({ selectedPrice });
       expect(context.questions).toEqual(questions);
```
alibombaye, line 97: Just wondering if we need all that

alibombaye, line 97: If the questions are getting copied over from the manifest then think it's fair to check the context.questions equals manifest.questions. That way we can remove the `const questions` object

app/pages/sections/catalogue-solutions/order-item/contextCreator.test.js
```diff
@@ -130,4 +138,77 @@ describe('catalogue-solutions order-item contextCreator', () => {
...
   });
+  describe('getErrorContext', () => {
+    const mockValidationErrors = [{
+      field: 'quantity',
+      id: 'quantityRequired',
+    },
+    {
+      field: 'price',
+      id: 'priceRequired',
+    }];
```
alibombaye, line 149: a blank line after this please. also the contents of these objects can be inlined if you want to save on some space
```
{ field: 'quantity', id: 'quantityRequired' }
```

app/pages/sections/catalogue-solutions/order-item/contextCreator.test.js
```diff
@@ -130,4 +138,77 @@ describe('catalogue-solutions order-item contextCreator', () => {
...
+        odsCode: 'ods-code',
+        selectedPrice,
+        validationErrors: mockValidationErrors,
+      };
+
+      getErrorContext(params);
+      expect(errorContext.getSectionErrorContext.mock.calls.length).toEqual(1);
+    });
+
+    it('should call add error message to the table data', async () => {
```
alibombaye, line 197: think this should be `'should add error message to the table data'` if I'm understanding it correctly

app/pages/sections/catalogue-solutions/order-item/controller.js
```diff
@@ -22,16 +22,56 @@ export const getSelectedPrice = async ({ selectedPriceId, accessToken }) => {
...
+      id: 'quantityRequired',
+    });
+  }
+
+  if (!data.quantity || isNaN(data.quantity)) {
+    errors.push({
+      field: 'quantity',
+      id: 'numericQuantityRequired',
+    });
+  }
```
alibombaye, line 49: is this right? If I put in `blah` into quantity I am going to get Quantity is required rather that Quantity must be a number

alibombaye: ignore me..  I missed the first bit.

app/pages/sections/catalogue-solutions/order-item/contextCreator.js
samMcCarthyBJSS: becuase this is just for the post with the extra data

app/pages/sections/catalogue-solutions/order-item/controller.js
samMcCarthyBJSS: image

app/pages/sections/catalogue-solutions/order-item/controller.test.js
```diff
@@ -101,4 +84,96 @@ describe('catalogue-solutions order-item controller', () => {
...
+      it('should return an array of one validation error and success as false if empty string for quantity is passed in', () => {
+        const data = {
+          quantity: '',
+          price: '1.5',
+        };
+
+        const response = validateOrderItemForm({ data });
+
+        expect(response.success).toEqual(false);
+        expect(response.errors).toEqual([quantityRequired, numericalQuantity]);
```
alibombaye, line 129: should it just be one error? quantityIsRequired. Don't check the numerical until something is provided

app/pages/sections/catalogue-solutions/order-item/controller.test.js
```diff
@@ -101,4 +84,96 @@ describe('catalogue-solutions order-item controller', () => {
...
+      it('should return an array of one validation error and success as false if empty string for price is passed in', () => {
+        const data = {
+          quantity: '1',
+          price: '',
+        };
+
+        const response = validateOrderItemForm({ data });
+
+        expect(response.success).toEqual(false);
+        expect(response.errors).toEqual([priceRequired, numericalPrice]);
```
alibombaye, line 153: Same as above. Just one error saying the price is required

app/pages/sections/catalogue-solutions/order-item/controller.test.js
```diff
@@ -101,4 +84,96 @@ describe('catalogue-solutions order-item controller', () => {
...
+        expect(response.errors).toEqual([numericalPrice]);
+      });
+
+      it('should return a validation error if all values are undefined', () => {
+        const data = {};
+
+        const response = validateOrderItemForm({ data });
+
+        expect(response.errors).toEqual(
+          [quantityRequired, numericalQuantity, priceRequired, numericalPrice],
```
alibombaye, line 174: this should just be quantity and price required.

But really nice tests.. really made it clear what the function is returning back

app/pages/sections/catalogue-solutions/order-item/controller.js
```diff
@@ -22,16 +22,56 @@ export const getSelectedPrice = async ({ selectedPriceId, accessToken }) => {
...
+      id: 'priceRequired',
+    });
+  }
+
+  if (!data.price || isNaN(data.price)) {
+    errors.push({
+      field: 'price',
+      id: 'numericPriceRequired',
+    });
+  }
```
alibombaye, line 63: are we missing a validation to check an estimation period is selected?

app/pages/sections/catalogue-solutions/order-item/ui-tests/general.ui.test.js
```diff
@@ -48,25 +60,61 @@ const selectedPrice = {
...
+  currencyCode: 'GBP',
+  itemUnit: {
+    name: 'patient',
+    description: 'per patient',
+  },
+  timeUnit: {
+    name: 'year',
+    description: 'per year',
+  },
+};
```
alibombaye, line 76: mmm I don't think using this as the returned data makes this an accurate ui test. Populating the form with blank data and then clicking submit.  BAPI should most likely return these values and the user will then delete the contents of these from the fields and then click submit. I think commencement date ui test is an example of where we delete data in the form and click submit

app/helpers/questionExtractor.test.js
```diff
@@ -0,0 +1,19 @@
...
+  ],
+};
+describe('questionExtractor', () => {
+  it('returns the correct question', () => {
+    expect(questionExtractor('test-id', manifest)).toEqual(manifest.questions[0]);
+  });
+
+  it('does not return a question if there is no match', () => {
+    expect(questionExtractor('wrong-id', manifest)).toEqual(undefined);
+  });
```
alibombaye, line 18: perfect

app/pages/sections/catalogue-solutions/order-item/ui-tests/general.ui.test.js
samMcCarthyBJSS: but then the end result is the same that nothin is populated in the box before the submit button is clicked?

app/pages/sections/catalogue-solutions/order-item/contextCreator.js
alibombaye: mm ok but won't it be `undefined` if it's not provided? don't think defaulting it to `null` adds any value

samMcCarthyBJSS: yeah true will get rid

app/pages/sections/catalogue-solutions/order-item/ui-tests/general.ui.test.js
alibombaye: I totally agree but I don't think is the right way to mimic that. We should reflect how that app is going behave and the users interactions with it. I think it's just the price that is populated and can be removed before submitting. 

This way if someone else comes to look at these we know we are always expecting the price to be provided to us and it's possible for this value to be deleted by the user and submitted

samMcCarthyBJSS: yeah that's fair will delete the price instead then

PR #178, opened 2020-06-30
Feature/7731 show recurring cost row

bjss-chris:
Story: [AB#7731](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/7731)
Task: 7884

image


app/pages/preview/contextCreator.js
```diff
@@ -58,7 +58,73 @@ const generateCallOffAndSupplierDetailsTable = ({
...
 
-export const getContext = ({ orderId, orderData }) => ({
+const generateRecurringCostDetailsTable = ({
+  recurringCostTable, recurringCostItems = [], serviceRecipients = {},
+}) => {
+  const items = recurringCostItems.map((item) => {
+    const classes = 'nhsuk-u-font-size-14';
+    const columns = [];
+
+    if (!Object.prototype.hasOwnProperty.call(serviceRecipients, item.serviceRecipientsOdsCode)) {
```
alibombaye, line 68: think you can shorthand this too

`if (!serviceRecipients[item.serviceRecipientsOdsCode])`

app/pages/preview/contextCreator.js
```diff
@@ -58,7 +58,73 @@ const generateCallOffAndSupplierDetailsTable = ({
...
+      throw new Error(`service recipient ${item.serviceRecipientsOdsCode} not found`);
```
alibombaye, line 69: The pattern we have used in the app for throwing errors is something like this

```
logger.error(`blah blah blah`);
throw new Error();
```

app/pages/preview/helpers/getServiceRecipients.js
```diff
@@ -0,0 +1,9 @@
+export const getServiceRecipients = (serviceRecipients = []) => {
```
alibombaye, line 1: just on the name of this function. Any reason you went for `getServiceRecipients`? we normally use the `get` prefix when it's going to call the api to get some data. Do you think we could rename this function to something else?

bjss-chris: I wasn't aware that you reserved the `get` prefix for API calls and I just thought that `getServiceRecipients` was more concise than `createServiceRecipientsDict`. However, I'll rename to the latter.

PR #180, opened 2020-07-01
Feature/5127 show totals

alibombaye:
Story [AB#5127](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/5127)
Task [AB#8070](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/8070)

Screenshot 2020-07-01 at 17 05 33

Showing totals on the orderSummary page

### DIFF
--- a/app/helpers/priceFormatter.js
+++ b/app/helpers/priceFormatter.js
@@ -1,0 +1,9 @@
+export const formatPrice = (priceValue) => {
+  const formattedPrice = priceValue.toLocaleString(undefined, {
+    minimumFractionDigits: 3, maximumFractionDigits: 3,
+  });
+
+  const truncatedTo2dp = formattedPrice.substring(0, formattedPrice.indexOf('.') + 3);
+
+  return truncatedTo2dp;
+};
--- a/app/helpers/priceFormatter.test.js
+++ b/app/helpers/priceFormatter.test.js
@@ -1,0 +1,14 @@
+import { formatPrice } from './priceFormatter';
+
+describe('formatPrice', () => {
+  it.each`
+    priceValue   | expected
+    ${10}        | ${'10.00'}
+    ${1000}      | ${'1,000.00'}
+    ${999.999}   | ${'999.99'}
+    ${1234.560}  | ${'1,234.56'}
+    ${12.011}    | ${'12.01'}
+  `('formatPrice $priceValue should be formatted to $expected', ({ priceValue, expected }) => {
+  expect(formatPrice(priceValue)).toEqual(expected);
+});
+});
--- a/app/pages/preview/contextCreator.test.js
+++ b/app/pages/preview/contextCreator.test.js
@@ -188,17 +188,13 @@ describe('order summary preview contextCreator', () => {
           items: [
             [
               {
-                data: '',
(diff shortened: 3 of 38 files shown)


app/helpers/priceFormatter.test.js
```diff
@@ -0,0 +1,14 @@
+import { formatPrice } from './priceFormatter';
+
+describe('formatPrice', () => {
+  it.each`
+    priceValue   | expected
+    ${10}        | ${'10.00'}
```
samMcCarthyBJSS, line 6: shouldn't the price be 3 decimal places?

app/helpers/priceFormatter.js
```diff
@@ -0,0 +1,9 @@
+export const formatPrice = (priceValue) => {
+  const formattedPrice = priceValue.toLocaleString(undefined, {
+    minimumFractionDigits: 3, maximumFractionDigits: 3,
+  });
+
+  const truncatedTo2dp = formattedPrice.substring(0, formattedPrice.indexOf('.') + 3);
```
samMcCarthyBJSS, line 6: do we need to keep waht the original value is? or else we're losing data

app/helpers/priceFormatter.test.js
alibombaye: This function is to format the price for display purposes. which is to always round down to the nearest 2 decimal place number.  We may get the price come back to us from the API in 3 decimal places

app/helpers/priceFormatter.js
alibombaye: mmm good shout but I don't think so.. This data will come from ordapi which will have the original value and for the preview we never update the order so it will just be for display purposes only

app/pages/preview/contextCreator.test.js
```diff
@@ -177,6 +175,67 @@ describe('order summary preview contextCreator', () => {
...
+            ],
+          ],
+        },
+      };
+
+      const context = getContext({ orderId: 'order-1', orderData: mockOrderData });
+      expect(context.oneOffCostTotalsTable).toEqual(expectedContext.oneOffCostTotalsTable);
+    });
+
+    it('should return the oneOffCostTotalsTable with items and the total cost value set to when provided', () => {
```
bjss-chris, line 209: Set to [something missing here]?

app/pages/preview/contextCreator.test.js
```diff
@@ -260,5 +319,145 @@ describe('order summary preview contextCreator', () => {
...
+          ],
+        },
+      };
+
+      const context = getContext({ orderId: 'order-1', orderData: mockOrderData });
+
+      expect(context.recurringCostTotalsTable).toEqual(expectedContext.recurringCostTotalsTable);
+    });
+
+    it('should return the recurringCostTotalsTable with items and the total cost value set to when provided', () => {
```
bjss-chris, line 390: Set to [something missing here]?

app/pages/preview/contextCreator.test.js
alibombaye: good spot. I've updated these.

PR #190, opened 2020-07-06
Feature/7848 create catalogue solution

stuart-bass-bjss:
[AB#8119](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/8119) - Edit a single Catalogue Solution - Flat List Price - Variable (on-demand) - splitx2
[AB#7848](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/7848) - FE - OF - Call POST to create catalogueOrderItem


app/pages/sections/catalogue-solutions/order-item/controller.js
```diff
@@ -1,12 +1,33 @@
...
-import { getData } from 'buying-catalogue-library';
+import { getData, postData } from 'buying-catalogue-library';
 import { getContext, getErrorContext } from './contextCreator';
 import { logger } from '../../../../logger';
 import { getEndpoint } from '../../../../endpoints';
 import commonManifest from './commonManifest.json';
 import { getSelectedPriceManifest } from './manifestProvider';
 import { getDateErrors } from '../../../../helpers/getDateErrors';
 
+const extractDeliveryDate = (detail) => {
```
alibombaye, line 10: So we do something similar in the commencementDate section where we pass in the date from FE and convert it to a date we can pass to the API. How about creating a generic function and moving this into the helpers folder so it can be used by both?

app/pages/sections/catalogue-solutions/order-item/controller.js
```diff
@@ -156,7 +177,22 @@ export const validateOrderItemForm = ({ data, selectedPrice }) => {
 export const getSolution = async ({ solutionId, accessToken }) => {
   const endpoint = getEndpoint({ api: 'bapi', endpointLocator: 'getSolution', options: { solutionId } });
   const solutionData = await getData({ endpoint, accessToken, logger });
-  logger.info(`Retrived solution data from BAPI for ${solutionId}`);
+  logger.info(`Retrieved solution data from BAPI for ${solutionId}`);
 
   return solutionData;
 };
+
+export const postSolution = async ({
```
alibombaye, line 185: If you could rename this to postSolutionOrderItem to make it little more explicit that is what is going on here

app/pages/sections/catalogue-solutions/catalogue-solutions/controller.js
```diff
@@ -17,7 +17,7 @@ export const getCatalogueSolutionsPageContext = async ({ orderId, accessToken })
   return getContext({
     orderId,
     orderDescription: orderDescriptionData ? orderDescriptionData.description : '',
-    orderItems: addedSolutionsData.orderItems,
+    orderItems: addedSolutionsData.catalogueSolutions,
```
alibombaye, line 20: just wondering why this was changed?

app/pages/sections/catalogue-solutions/order-item/controller.test.js
```diff
@@ -31,6 +40,16 @@ const selectedPrice = {
   price: 0.1,
 };
 
+const detail = {
```
alibombaye, line 43: rather than declare this at the top can you just declare it in test it is being used

app/pages/sections/catalogue-solutions/order-item/controller.test.js
```diff
@@ -19,6 +19,15 @@ jest.mock('./manifestProvider', () => ({
...
 
+const serviceRecipient = {
+  name: 'Recipient 1',
+  odsCode: 'ods1',
+};
+const solution = {
+  id: 'solutionId1',
+  name: 'Solution 1',
+};
+
```
alibombaye, line 30: same with the above 2 consts if you could move them to the test that uses them. I think it's only being used by that one test

app/pages/sections/catalogue-solutions/order-item/controller.test.js
```diff
@@ -336,4 +355,42 @@ describe('catalogue-solutions order-item controller', () => {
...
+          quantity: 1,
+          estimationPeriod: 'perMonth',
+          provisioningType: 'OnDemand',
+          type: 'flat',
+          currencyCode: 'GBP',
+          itemUnitModel: selectedPrice.itemUnit,
+          price: 500.49,
+        },
+        accessToken: 'access_token',
+        logger,
```
alibombaye, line 392: 👍 nice

app/pages/sections/catalogue-solutions/routes.js
```diff
@@ -74,16 +75,35 @@ export const catalogueSolutionsRoutes = (authProvider, addContext, sessionManage
...
     if (response.success) {
+      const selectedSolutionId = sessionManager.getFromSession({ req, key: 'selectedSolutionId' });
+      const serviceRecipient = {
+        name: serviceRecipientName,
+        odsCode: selectedRecipientId,
+      };
+      const solution = {
+        id: selectedSolutionId,
+        name: solutionName,
+      };
```
alibombaye, line 94: I personally would prefer if we didn't do the above 2 in the routes. If you can move this to the controller or even in the formatPostData that would be nice

app/pages/sections/catalogue-solutions/order-item/controller.js
stuart-bass-bjss: done

app/pages/sections/catalogue-solutions/order-item/controller.js
stuart-bass-bjss: done

app/pages/sections/catalogue-solutions/catalogue-solutions/controller.js
stuart-bass-bjss: because the api doesn't currently support orderItems for catalogue solutions, it still uses the old route.

stuart-bass-bjss: It'll need changing once the api is changed

app/pages/sections/catalogue-solutions/order-item/controller.test.js
stuart-bass-bjss: done

app/pages/sections/catalogue-solutions/catalogue-solutions/controller.js
alibombaye: ah ok. good spot on that one 👍

app/pages/sections/catalogue-solutions/routes.js
stuart-bass-bjss: done

PR #216, opened 2020-07-10
Feature/5352 show no additional service page

sam-roberts-bjss:
Story: [AB#5352](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/5352)
Task: [AB#8168](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/8168)

Screenshot 2020 07 10 at 09 24

### DIFF
--- a/app/pages/sections/additional-services/select/additional-service/controller.test.js
+++ b/app/pages/sections/additional-services/select/additional-service/controller.test.js
@@ -68,6 +68,31 @@ describe('additional-services select-additional-service controller', () => {
         logger,
       });
     });
+
+    it('should return empty list when getData returns undefined', async () => {
+      getData
+        .mockResolvedValueOnce({ additionalServices: undefined });
+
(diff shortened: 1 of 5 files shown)


app/pages/sections/additional-services/select/additional-service/controller.test.js
```diff
@@ -68,6 +68,31 @@ describe('additional-services select-additional-service controller', () => {
         logger,
       });
     });
+
+    it('should return empty list when getData returns undefined', async () => {
```
stuart-bass-bjss, line 72: what if it's returned not as an array, but something else? can it be made resilient to object type change?

sam-roberts-bjss: Apologies, not sure what you mean?

PR #227, opened 2020-07-13
get recipients

samMcCarthyBJSS:
Story [AB#7265](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/7265)
Task [AB#8326](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/8326)

### DIFF
--- a/app/pages/sections/additional-services/select/routes.test.js
+++ b/app/pages/sections/additional-services/select/routes.test.js
@@ -15,7 +15,7 @@ import * as additionalServicePriceController from './price/controller';
 import { App } from '../../../../app';
 import { routes } from '../../../../routes';
 import { baseUrl } from '../../../../config';
-import * as getRecipientsHelper from '../../../../helpers/api/ordapi/getRecipients';
+import { getRecipients } from '../../../../helpers/api/ordapi/getRecipients';
 import * as routerHelper from '../../../../helpers/routes/routerHelper';
 
 jest.mock('../../../../logger');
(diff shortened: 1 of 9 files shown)


app/pages/sections/additional-services/select/routes.test.js
```diff
@@ -343,7 +343,7 @@ describe('additional-services select routes', () => {
     ));
 
     it('should return the additional-services select recipient page if authorised', async () => {
-      getSolutionHelper.getSolution = jest.fn()
+      getRecipientsHelper.getRecipients = jest.fn()
```
alibombaye, line 346: I've recently discovered a cleaner way of mocking a file that has a single function.

in the import leave this as;
`import { getRecipients } from '../../../../helpers/api/ordapi/getRecipients';`

add a jest.mock under the imports
`jest.mock('../../../../helpers/api/ordapi/getRecipients');`

Then when we need to mock it we can simply do
`getRecipients.mockResolvedValue({});`

alibombaye: saves us having to alias it as something.

app/pages/sections/catalogue-solutions/select/routes.test.js
```diff
@@ -15,6 +15,7 @@ import * as selectRecipientController from './recipient/controller';
 import { App } from '../../../../app';
 import { routes } from '../../../../routes';
 import { baseUrl } from '../../../../config';
+import * as getRecipientsHelper from '../../../../helpers/api/ordapi/getRecipients';
```
alibombaye, line 18: as above. See if we can just mock function and not have to alias it.

PR #240, opened 2020-07-16
Feature/7263 update solution order items endpoint

sam-roberts-bjss:
Story: [AB#7263](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/7263)
Task: [AB#8320](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/8320)

### DIFF
--- a/app/helpers/api/ordapi/getOrderItems.js
+++ b/app/helpers/api/ordapi/getOrderItems.js
@@ -5,7 +5,7 @@ import { getEndpoint } from '../../../endpoints';
 export const getOrderItems = async ({ orderId, catalogueItemType, accessToken }) => {
   const endpoint = getEndpoint({ api: 'ordapi', endpointLocator: 'getOrderItems', options: { orderId, catalogueItemType } });
   const orderItems = await getData({ endpoint, accessToken, logger });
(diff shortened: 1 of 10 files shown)


app/helpers/api/ordapi/getOrderItems.js
```diff
@@ -0,0 +1,11 @@
+import { getData } from 'buying-catalogue-library';
+import { logger } from '../../../logger';
+import { getEndpoint } from '../../../endpoints';
+
+export const getOrderItems = async ({ orderId, catalogueItemType, accessToken }) => {
+  const endpoint = getEndpoint({ api: 'ordapi', endpointLocator: 'getOrderItems', options: { orderId, catalogueItemType } });
```
alibombaye, line 6: I would do what we chatted about the other day and start moving away from using this and doing what we do in endpoints and in here so it can be tested too.

sam-roberts-bjss: Yeah OK. Will give that a go 👍

PR #248, opened 2020-07-17
Feature/7580 update solution get endpoint

alibombaye:
Story [AB#7580](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/7580)
Task [AB#8312](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/8312)

Refactored catalogue-solutions to use new endpoint to getOrderItem details


app/pages/sections/order-items/catalogue-solutions/select/routes.js
```diff
@@ -141,12 +150,11 @@ export const catalogueSolutionsSelectRoutes = (authProvider, addContext, session
       return res.redirect(`${config.baseUrl}/organisation/${orderId}/catalogue-solutions/neworderitem`);
     }
 
-    const solutionId = sessionManager.getFromSession({ req, key: 'selectedSolutionId' });
-    const solutionData = await getSolution({ solutionId });
+    const solutionName = sessionManager.getFromSession({ req, key: 'selectedItemName' });
```
sam-roberts-bjss, line 153: Was thinking the variable name could match the key but due to the context of the test this is fine 👍

app/helpers/routes/findSelectedCatalogueItemInSession.js
```diff
@@ -0,0 +1,21 @@
+import { logger } from '../../logger';
+
+const findByType = {
+  additionalServices: catalogueItem => catalogueItem.additionalServiceId,
```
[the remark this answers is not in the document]
alibombaye: cool just to be clear this endpoint `/api/v1/catalogue-items`
returns the id as `catalogueItemId` rather than just `id` is that right?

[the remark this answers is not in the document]
sam-roberts-bjss: Correct 👍

app/pages/sections/order-items/catalogue-solutions/select/routes.js
alibombaye: I'll do this one later. Will leave for now but it's a valid point

app/helpers/api/bapi/getCatalogueItems.js
```diff
@@ -0,0 +1,15 @@
+import { getData } from 'buying-catalogue-library';
+import { logger } from '../../../logger';
+import { solutionsApiUrl } from '../../../config';
+
+const getCatalogueItemsEndpoint = (supplierId, catalogueItemType) => (
+  `${solutionsApiUrl}/api/v1/catalogue-items?supplierId=${supplierId}&catalogueItemType=${catalogueItemType}`
```
sam-roberts-bjss, line 6: Is there ever a case where you don't need to send a supplier ID or a catalogue item type?

app/helpers/api/bapi/getCatalogueItems.js
```diff
@@ -0,0 +1,15 @@
...
+);
+
+export const getCatalogueItems = async ({ supplierId, catalogueItemType, accessToken }) => {
+  const endpoint = getCatalogueItemsEndpoint(supplierId, catalogueItemType);
+  const catalogueItems = await getData({ endpoint, accessToken, logger });
+  logger.info(`Found ${catalogueItems.length} item(s) for supplier "${supplierId}".`);
```
sam-roberts-bjss, line 12: Very minor, could you get undefined back from the API?

app/helpers/api/bapi/getCatalogueItems.test.js
```diff
@@ -0,0 +1,25 @@
...
+    getData
+      .mockResolvedValueOnce({ data: {} });
+
+    await getCatalogueItems({ supplierId: 'supp-1', catalogueItemType: 'Solution', accessToken: 'access_token' });
+    expect(getData.mock.calls.length).toEqual(1);
+    expect(getData).toHaveBeenCalledWith({
+      endpoint: `${solutionsApiUrl}/api/v1/catalogue-items?supplierId=supp-1&catalogueItemType=Solution`,
+      accessToken: 'access_token',
+      logger,
+    });
```
sam-roberts-bjss, line 23: Need a test to check the response data or not?

app/helpers/api/bapi/getCatalogueItems.js
alibombaye: I was thinking this and if the only other user of this endpoint will be associated services then the endpoint will be the same.

So I think we should leave this as is for now. Since if them values are not provided then it's not a valid endpoint for the FE

app/helpers/api/bapi/getCatalogueItems.js
alibombaye: that's a good point. I'll update 👍

app/helpers/api/bapi/getCatalogueItems.test.js
alibombaye: yeah I'll add one 👍

app/helpers/routes/findSelectedCatalogueItemInSession.js
```diff
@@ -0,0 +1,23 @@
...
+}) => {
+  const additionalServices = sessionManager.getFromSession({ req, key: catalogueItemsKey });
+  const findCallback = catalogueItem => (
+    findByType[catalogueItemsKey](catalogueItem) === selectedItemId
+  );
+  const selectedItem = additionalServices.find(findCallback);
+
+  if (!selectedItem) {
+    logger.error(`Unable to find selected item ${selectedItemId} in session`);
+    throw new Error();
```
sam-roberts-bjss, line 19: Just suggestions to think about not saying change it...
Is it really an exceptional case to not find the selected item in session? How does the calling code know what to do with this error? Can it recover from it?

My thinking here is I want to check an item is in session & if not then I would then go to the API to retrieve it. An undefined check could satisfy this.

alibombaye: this one is a little tricky. We would have to know how to handle this and which endpoint to call again to get the list of catalogueItems again. I think what this is doing here is fine for now. The session has expired or for whatever reason it can't find the selectedItem it will show the general error page and they'll have to start the journey again. To try and mitigate this would add a fair bit of complexity to the code.

PR #259, opened 2020-07-20
Feature/7580 update solution post endpoint

alibombaye:
Story [AB#7580](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/7580)
Task [AB#8405](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/8405) [AB#8406](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/8406)


app/helpers/controllers/saveOrderItem.js
```diff
@@ -0,0 +1,45 @@
...
+      })
+      : await putOrderItem({
+        accessToken,
+        orderId,
+        orderItemId,
+        formData,
+      });
+    return response;
+  } catch (err) {
+    if (err.response.status === 400 && err.response.data && err.response.data.errors) {
```
stuart-bass-bjss, line 38: I don't think you should be reshaping the response like this if there are errors. router.post('/:orderItemId', for additional services for example expect the errors to be in .errors

app/helpers/controllers/saveOrderItem.js
```diff
@@ -0,0 +1,45 @@
...
+      logger.info(`Validation errors returned from the API ${JSON.stringify(err.response.data.errors)}`);
+      return err.response.data;
+    }
+    logger.error(`Error saving order item for ${itemName} and ${serviceRecipientName} for order id: ${orderId}`);
+    throw new Error();
```
stuart-bass-bjss, line 43: The error makes it all the way out of the routes into the generic error handler. Wouldn't it be better to put the error message in the error class, so that the logging, etc gets handled by the generic handler?

alibombaye: We also do that in `app/helpers/routes/routerHelper.js` and the `withCatch` function. The generic error handler there will print the full stack trace. This line here gives a specific log error message for this context. 

We can make the errorHandler do a better job at perhaps pass this message through so the generic handler can log it there. But that will involve bit of re-work across the app to do that.

app/helpers/controllers/saveOrderItem.js
alibombaye: I'm not sure what you mean here. But we are not reshaping the error response. It's just how the error response is returned to us from the api and axios. In this situation we want to handle when we get a 400 with errors back from the api. 

This is a pattern we've used for error handling the API validation errors through out the app so far. We can have a think about how we can do this better.

stuart-bass-bjss: ah, OK, I misread it, I thought it was returning just the array of errors, not { errors: [] }, which would completely mismatch.
You next layer up tests though are expecting { success: false, errors: [{}] }), which this function won't produce. I think you are getting away with it as the absence of success evaluates to false.

stuart-bass-bjss: Can you please change it to include success:false?

app/helpers/controllers/saveOrderItem.js
stuart-bass-bjss: It would be better to combine the error message with the stack trace, so they appear together in a single message. If the system is under any load / once we have splunk in place, identifying the association of adjacent messages will be a lot harder. Can you please add a tech debt ticket to look at it?

alibombaye: tech debt ticket #8537 created

PR #268, opened 2020-07-21
0 value price fix

alibombaye:
Story [AB#8534](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/8534)
Task [AB#8596](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/8596)

### DIFF
--- a/app/helpers/contextCreators/generateAddPriceTable.test.js
+++ b/app/helpers/contextCreators/generateAddPriceTable.test.js
@@ -59,7 +59,7 @@ describe('generateAddPriceTable', () => {
     };
 
     const generatedAddPriceTable = generateAddPriceTable({
-      addPriceTable, price: '1892.23', itemUnitDescription: 'per patient', timeUnitDescription: 'per year',
+      addPriceTable, price: 1892.23, itemUnitDescription: 'per patient', timeUnitDescription: 'per year',
     });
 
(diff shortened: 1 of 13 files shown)


app/helpers/contextCreators/generateAddPriceTable.test.js
```diff
@@ -59,7 +59,7 @@ describe('generateAddPriceTable', () => {
     };
 
     const generatedAddPriceTable = generateAddPriceTable({
-      addPriceTable, price: '1892.23', itemUnitDescription: 'per patient', timeUnitDescription: 'per year',
+      addPriceTable, price: 1892.23, itemUnitDescription: 'per patient', timeUnitDescription: 'per year',
```
sam-roberts-bjss, line 62: Does this test represent data from BE, hence why it is a number & not a string?

alibombaye: yeah it comes from the BE as a number. So updated the test data to reflect that

alibombaye:
> 👍 Changes look good. Wondering whether or not you need a specific test for 0?

good point. I'll add one 👍

PR #287, opened 2020-07-24
update session keys and clear session on succesfull post

alibombaye:
Story [AB#8737](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/8737)
Task [AB#8744](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/8744)

### DIFF
--- a/app/pages/sections/order-items/additional-services/routes.js
+++ b/app/pages/sections/order-items/additional-services/routes.js
@@ -104,6 +104,12 @@ export const additionalServicesRoutes = (authProvider, addContext, sessionManage
       });
 
       if (apiResponse.success) {
+        sessionManager.clearFromSession({
+          req,
+          keys: [
+            'selectedItemId', 'selectedItemName', 'selectedRecipientId', 'selectedRecipientName', 'selectedPriceId', 'selectedCatalogueSolutionId',
+          ],
+        });
         logger.info('Redirecting to the additional-services main page');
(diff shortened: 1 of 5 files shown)


app/pages/sections/order-items/additional-services/routes.js
```diff
@@ -104,6 +104,12 @@ export const additionalServicesRoutes = (authProvider, addContext, sessionManage
       });
 
       if (apiResponse.success) {
+        sessionManager.clearFromSession({
+          req,
+          keys: [
+            'selectedItemId', 'selectedItemName', 'selectedRecipientId', 'selectedRecipientName', 'selectedPriceId', 'selectedCatalogueSolutionId',
+          ],
+        });
```
bjss-chris, line 112: Is it worth moving this into a separate function that we can reuse across additional services, associated services and catalogue solutions?

alibombaye: Good shout and I was considering this also. However, I didn't want to do anything too overdo this one just yet. I think we need to discuss about the best way to handle our sessions in general and where to tackle them.

bjss-chris: Ah, OK. I guess we can come back to it later as tech debt then.

PR #296, opened 2020-07-28
Feature/5437 submitting funding source bug

alibombaye:
Story [AB#5437](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/5437)
Task [AB#8783](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/8783)

### DIFF
--- a/app/helpers/api/ordapi/putFundingSource.js
+++ b/app/helpers/api/ordapi/putFundingSource.js
@@ -3,7 +3,7 @@ import { logger } from '../../../logger';
 import { orderApiUrl } from '../../../config';
 
 const formatPutData = ({ fundingSource }) => ({
-  onlyGMS: fundingSource,
+  onlyGMS: fundingSource === 'true',
 });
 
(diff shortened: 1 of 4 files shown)


app/helpers/api/ordapi/putFundingSource.js
```diff
@@ -3,7 +3,7 @@ import { logger } from '../../../logger';
 import { orderApiUrl } from '../../../config';
 
 const formatPutData = ({ fundingSource }) => ({
-  onlyGMS: fundingSource,
+  onlyGMS: fundingSource === 'true',
```
sam-roberts-bjss, line 6: Assume you don't need to worry about case here?

alibombaye: We should be good here to not worry about the case

PR #304, opened 2020-07-30
Add getOrderDescription function

bjss-chris:
Activity: [AB#8866](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/8866)
Task: [AB#8867](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/8867)


app/helpers/order/order.js
```diff
@@ -0,0 +1,21 @@
+import { getOrderDescription as getOrderDescriptionFromApi } from '../api/ordapi/getOrderDescription';
```
alibombaye: I think we should call the name of the file `getOrderDescription` too and perhaps it should live in the `helper/routes` folder.

bjss-chris: Yeah, noticed that and just renamed it!

bjss-chris: I created the `order` folder as I thought we may end up adding functions along the lines of `getSelectedCatalogueSolution` etc. at some point (as we talked about in our session session the other day) and I thought that fit better than routes as in terms of concepts all this stuff relates to an order.

Personally, I think I'd rather leave it where it is but you know this code base better than I do so I don't mind moving it. What do you think?

alibombaye: I understand where you are coming from with naming the folder `order`. However, the helper folder is organised by function rather than domain at the moment. I want to avoid having a mixture or both since that might confuse us to where things should live. But if we feel that structuring by domain may be better then let's create a tech debt to tackle this.

alibombaye: Or maybe these are best placed within the file where the call to get the orderDescription from ordapi is done? what do you think?

bjss-chris: I've gone with the routes directory for now. I thought it might be better to keep the stuff in api for pure API stuff, if that works for you too?

bjss-chris:
> 👍 other than the name and location. I think we also suggested as part of this task was to have a dedicated function for getting order description from the session.
> 
> Perhaps a session manager helper file that has these dedicated functions and wraps stuff like this
> `sessionManager.getFromSession({ req, key });`
> 
> Or if there is a neater way you can think of.

I've added a separate function for retrieving data from session. I don't think it really adds much value right now but hopefully it might do as we do additional work refactoring the session management.

app/helpers/order/order.js
alibombaye: yeah that's fine. I think it's worth revisiting this at some point though.

app/helpers/routes/sessionHelper.js
```diff
@@ -0,0 +1,15 @@
+export const getFromSessionOrApi = async ({
```
alibombaye, line 1: oooh I like this. Something we'll be able to start using for calls that follow this pattern 👍

PR #308, opened 2020-07-31
Feature/5145 call ordapi description

samMcCarthyBJSS:
Story [AB#5145](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/5145)
Task [AB#8863](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/8863)

### DIFF
--- a/app/pages/sections/routes.js
+++ b/app/pages/sections/routes.js
@@ -20,7 +20,7 @@ import { additionalServicesRoutes } from './order-items/additional-services/rout
 import { associatedServicesRoutes } from './order-items/associated-services/routes';
 import { getFundingSource } from '../../helpers/api/ordapi/getFundingSource';
 import { putFundingSource } from '../../helpers/api/ordapi/putFundingSource';
-import { getOrderDescription } from '../../helpers/api/ordapi/getOrderDescription';
+import { getOrderDescription } from '../../helpers/routes/getOrderDescription';
 
 const router = express.Router({ mergeParams: true });
 
@@ -49,10 +49,7 @@ export const sectionRoutes = (authProvider, addContext, sessionManager) => {
       accessToken,
     });
 
-    if (response.success) {
-      sessionManager.saveToSession({ req, key: 'description', value: req.body.description.trim() });
-      return res.redirect(`${config.baseUrl}/organisation/${response.orderId}`);
-    }
+    if (response.success) return res.redirect(`${config.baseUrl}/organisation/${response.orderId}`);
 
     const context = await getDescriptionErrorContext({
       validationErrors: response.errors,
--- a/app/pages/sections/routes.js
+++ b/app/pages/sections/routes.js
@@ -196,7 +196,7 @@ export const sectionRoutes = (authProvider, addContext, sessionManager) => {
     const { orderId } = req.params;
     const accessToken = extractAccessToken({ req, tokenType: 'access' });
     await getFundingSource({ orderId, accessToken });
-    getOrderDescription({
+    await getOrderDescription({
       req,
       sessionManager,
       accessToken,
--- a/app/pages/sections/routes.js
+++ b/app/pages/sections/routes.js
@@ -20,6 +20,7 @@ import { additionalServicesRoutes } from './order-items/additional-services/rout
 import { associatedServicesRoutes } from './order-items/associated-services/routes';
 import { getFundingSource } from '../../helpers/api/ordapi/getFundingSource';
 import { putFundingSource } from '../../helpers/api/ordapi/putFundingSource';
+import { getOrderDescription } from '../../helpers/api/ordapi/getOrderDescription';
 
 const router = express.Router({ mergeParams: true });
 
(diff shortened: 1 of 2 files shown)


app/pages/sections/routes.js
```diff
@@ -20,6 +20,7 @@ import { additionalServicesRoutes } from './order-items/additional-services/rout
 import { associatedServicesRoutes } from './order-items/associated-services/routes';
 import { getFundingSource } from '../../helpers/api/ordapi/getFundingSource';
 import { putFundingSource } from '../../helpers/api/ordapi/putFundingSource';
+import { getOrderDescription } from '../../helpers/api/ordapi/getOrderDescription';
```
bjss-chris, line 23: I think we probably want to use the `getOrderDescription` that I've added as part of activity 8866 here, so this should be:

```javascript
import { getOrderDescription } from '../../helpers/routes/getOrderDescription';
```

app/pages/sections/routes.js
```diff
@@ -41,7 +42,10 @@ export const sectionRoutes = (authProvider, addContext, sessionManager) => {
       accessToken,
     });
 
-    if (response.success) return res.redirect(`${config.baseUrl}/organisation/${response.orderId}`);
+    if (response.success) {
+      sessionManager.saveToSession({ req, key: 'description', value: req.body.description.trim() });
```
bjss-chris, line 46: I've created a `sessionKeys` object as part of task 8868 (PR currently awaiting review). Assuming we merge that PR before this one then I'd recommend we use that for the session key here.

It looks like we've used `orderDescription` as the session key elsewhere but that's irrelevant if we do what I've suggested above – unless there's a specific reason for the use of a different key here?

app/pages/sections/routes.js
```diff
@@ -184,6 +188,10 @@ export const sectionRoutes = (authProvider, addContext, sessionManager) => {
     const { orderId } = req.params;
     const accessToken = extractAccessToken({ req, tokenType: 'access' });
     await getFundingSource({ orderId, accessToken });
+    let description = sessionManager.getFromSession({ req, key: 'description' });
+    if (!description) {
+      description = await getOrderDescription({ orderId, accessToken });
+    }
```
bjss-chris, line 194: This can be replaced by a single call to the new getOrderDescription function. You'll have to pass some additional parameters though, `req` (instead of `orderId`), `sessionManager`, and `logger` if I remember correctly.

app/pages/sections/routes.test.js
```diff
@@ -18,11 +18,13 @@ import * as commencementDateController from './commencement-date/controller';
 import * as serviceRecipientsController from './service-recipients/controller';
 import { getFundingSource } from '../../helpers/api/ordapi/getFundingSource';
 import { putFundingSource } from '../../helpers/api/ordapi/putFundingSource';
+import { getOrderDescription } from '../../helpers/api/ordapi/getOrderDescription';
```
bjss-chris, line 21: As above, I think we probably want to use the getOrderDescription that I've added as part of activity 8866 here, so this should be:

```javascript
import { getOrderDescription } from '../../helpers/routes/getOrderDescription';
```

app/pages/sections/routes.test.js
```diff
@@ -18,11 +18,13 @@ import * as commencementDateController from './commencement-date/controller';
...
 import * as fundingSourceController from './funding-source/controller';
 
 jest.mock('../../logger');
 jest.mock('../../helpers/api/ordapi/getFundingSource');
 jest.mock('../../helpers/api/ordapi/putFundingSource');
+jest.mock('../../helpers/api/ordapi/getOrderDescription');
```
bjss-chris, line 27: As above, will become:

```javascript
jest.mock('../../helpers/routes/getOrderDescription');;
```

app/pages/sections/routes.js
```diff
@@ -49,10 +49,7 @@ export const sectionRoutes = (authProvider, addContext, sessionManager) => {
       accessToken,
     });
 
-    if (response.success) {
-      sessionManager.saveToSession({ req, key: 'description', value: req.body.description.trim() });
-      return res.redirect(`${config.baseUrl}/organisation/${response.orderId}`);
-    }
+    if (response.success) return res.redirect(`${config.baseUrl}/organisation/${response.orderId}`);
```
bjss-chris: Sorry, just thinking about this I think we do need to clear the order description from session here (on success) or update it as you had originally; otherwise, the call to `getOrderDescription` will retrieve the value we might previously have saved to session.

Once that's done, I think covers task 8869 that I was just about to pick up but we may as well do it as part of this change I guess.

samMcCarthyBJSS: should I save it as orderDescription or just description?

bjss-chris: You can use `sessionKeys.orderDescription` from `sessionHelper`.

app/pages/sections/routes.js
```diff
@@ -195,6 +196,12 @@ export const sectionRoutes = (authProvider, addContext, sessionManager) => {
     const { orderId } = req.params;
     const accessToken = extractAccessToken({ req, tokenType: 'access' });
     await getFundingSource({ orderId, accessToken });
+    getOrderDescription({
+      req,
+      sessionManager,
+      accessToken,
+      logger,
+    });
```
bjss-chris, line 204: Is there a reason why the results of `getFundingSource` and `getOrderDescription` aren't used?

`getOrderDescription` is async so will need an `await` too.

samMcCarthyBJSS: yeah the story is just for calling it
will use them in the next story for create basic page

PR #328, opened 2020-08-05
Feature/8870 clear session at start

alibombaye:
Story [AB#8870](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/8870)
Task [AB#8751](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/8751)

### DIFF
--- a/app/helpers/routes/sessionHelper.js
+++ b/app/helpers/routes/sessionHelper.js
@@ -34,3 +34,13 @@ export const getFromSessionOrApi = async ({
 
   return value;
 };
+
+export const clearSession = async ({
+  req,
+  sessionManager,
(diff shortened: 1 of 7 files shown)


alibombaye:
> There are a couple of session keys e.g. order description & funding source for the complete order confirmation page, will need to figure out where to clear them as I don't think the user will be able to access the dashboard after "completing".
> 
> Like the simplification though 💯

yeah that is a good spot. Didn't think of that. Hopefully it's just a case of calling that clearSession function whenever we need it.

PR #346, opened 2020-08-10
added incomplete check

samMcCarthyBJSS:
Story [AB#8904](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/8904)
Task [AB#9087](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/9087)


app/pages/summary/contextCreator.js
```diff
@@ -316,36 +316,40 @@ const generateOneOffCostDetailsTable = ({
...
+    }),
+    recurringCostTotalsTable: generateRecurringCostTotalsTable({
+      recurringCostTotalsTable: manifest.recurringCostTotalsTable,
+      recurringYearCost: orderData.totalRecurringCostPerYear,
+      recurringMonthCost: orderData.totalRecurringCostPerMonth,
+      ownershipCost: orderData.totalOwnershipCost,
+    }),
+    commencementDate: formatDate(orderData.commencementDate),
+    backLinkHref: `${baseUrl}/organisation/${orderId}`,
+  });
```
alibombaye, line 354: the use of manifest within here would also need changing. I think it might be better to import the `'./incomplete/manifest.json';` as incompleteOrderManifest.

Then keep `manifestFromStatus` as just manifest. Then you shouldn't have to change this all here.

PR #347, opened 2020-08-10
Feature/5287 resize buttons

alibombaye:
Story [AB#5287](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/5287)
Task [AB#9251](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/9251)

Buttons content changed rather than resized

Screenshot 2020-08-10 at 14 37 28


app/pages/delete-order/manifest.json
```diff
@@ -3,6 +3,10 @@
   "title": "Delete order {{orderId}}?",
   "description": "Confirm you want to delete this order.",
   "orderDescriptionTitle": "Order description",
-  "noButtonText": "No",
-  "yesButtonText": "Yes"
+  "noButton": {
```
sam-roberts-bjss, line 6: Does this need to be an object?

alibombaye: no it doesn't but I started with resizing and was about to put the width in there. But then the content got changed so didn't need the width. but I actually prefer it as an object. gives us bit more scope to add more config to the button.

PR #363, opened 2020-08-19
sort catalogue solutions

samMcCarthyBJSS:
Story [AB#9004](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/9004)
Task [AB#9318](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/9318)

image


app/helpers/api/bapi/getCatalogueItems.js
```diff
@@ -28,6 +28,7 @@ export const getCatalogueItems = async ({ supplierId, catalogueItemType }) => {
 
   logger.info(`Retrieving ${catalogueItemType} catalogue items from BAPI for supplier ${supplierId}`);
   const catalogueItems = await getData({ endpoint, logger });
+  catalogueItems.sort((a, b) => ((a.name > b.name) ? 1 : -1));
```
alibombaye, line 31: Move this to a little helper function within this file. Also can you lowerCase the name for comparing. Otherwise I think we might get something like this;

A Solution
B Solution 
a solution
b solution

Example of where we have done something similar
```
const sortItems = (items, propToSort) => (
  items.sort((itemA, itemB) => {
    const itemAPropValue = itemA[propToSort].toLowerCase();
    const itemBPropValue = itemB[propToSort].toLowerCase();

    if (itemAPropValue < itemBPropValue) return -1;
    if (itemAPropValue > itemBPropValue) return 1;
    return 0;
  })
);
```

app/helpers/api/bapi/getCatalogueItems.js
```diff
@@ -28,6 +29,7 @@ export const getCatalogueItems = async ({ supplierId, catalogueItemType }) => {
 
   logger.info(`Retrieving ${catalogueItemType} catalogue items from BAPI for supplier ${supplierId}`);
   const catalogueItems = await getData({ endpoint, logger });
+  sortItems(catalogueItems, 'name');
```
alibombaye, line 32: I would have thought here that it would return a new array that is sorted. But the test prove that it works as is 🤷

app/helpers/common/sortItems.test.js
```diff
@@ -0,0 +1,36 @@
+import { sortItems } from './sortItems';
+
+describe('sort service recipients by name', () => {
```
alibombaye, line 3: update test description `sortItems`

app/helpers/common/sortItems.test.js
```diff
@@ -0,0 +1,36 @@
...
+
+  it('should return a list of 1 item sorted by correct key', () => {
+    const toSortArray = [toSort1];
+
+    const sortedServiceRecipients = sortItems(toSortArray, 'sortBy');
+
+    expect(sortedServiceRecipients).toEqual(toSortArray);
+  });
+
+  it('should return a list of 3 service recipients sorted by name', () => {
```
alibombaye, line 27: update test description but these tests are nice 👍

app/helpers/common/sortItems.test.js
```diff
@@ -0,0 +1,36 @@
...
+    const sortedArray = [toSort1, toSort2, toSort3];
+
+    const toSortArray = [toSort2, toSort3, toSort1];
+
+    const sortedServiceRecipients = sortItems(toSortArray, 'sortBy');
```
alibombaye, line 32: change the const to `sortedItems`

app/helpers/api/bapi/getCatalogueItems.js
samMcCarthyBJSS: yeah it's cos js uses the same memory positions
was having a bit of test issue because of this earlier overwriting what the expected result should be

alibombaye: oh I don't like that. Can you take a copy of the array that is passed to sortItems and return the copied array that is sorted.

alibombaye: leave this as is. Deep cloning the array in sortItems could cause further issues.

alibombaye (approved):

PR #367, opened 2020-08-27
update route when only one price

samMcCarthyBJSS:
Story [AB#9228](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/9228)
Task [AB#9326](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/9326)


alibombaye (commented):
👍  code looks good. Is there a UI test too we can add to check this happens also?

alibombaye (approved):

PR #377, opened 2020-10-06
Fix default delivery date bugs

bjss-chris:
Fixes the following bugs:
- lowercase 'c' on continue button
- missing validation error message when date more than 183 weeks after commencement date
- missing validation error message when date before commencement date
- selected service recipients are not remembered when navigating back

Story: [AB#8952](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/8952)
Task: [AB#9982](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/9982)

### DIFF
--- a/app/helpers/api/ordapi/putPlannedDeliveryDate.js
+++ b/app/helpers/api/ordapi/putPlannedDeliveryDate.js
@@ -4,7 +4,7 @@ import { orderApiUrl } from '../../../config';
 import { extractDate } from '../../controllers/extractDate';
 
 const formatPutData = data => ({
-  deliveryDate: extractDate('plannedDeliveryDate', data),
+  deliveryDate: extractDate('deliveryDate', data),
 });
 
 const getPlannedDeliveryDateEndpoint = (orderId, catalogueItemId, priceId) => (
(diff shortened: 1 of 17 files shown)


samMcCarthyBJSS (commented):
Should the time difference bugs not have been done on the back end like they were for commencement date?

And what was the service recipients change for?

bjss-chris:
> Should the time difference bugs not have been done on the back end like they were for commencement date?

Are you referring to the validation error IDs? I considered whether to change the FE or the BE and I decided to go with the FE (even though it meant a few more changes) because the back-end currently meets the existing JSON contract. That uses `default-delivery-date` and `deliveryDate` so I didn't want to introduce references to `PlannedDeliveryDate`, which doesn't tie in to the contract (or anything else in the BE). I could've changed the contract but that would've meant changes to both the BE and FE.

> And what was the service recipients change for?

The service recipients change is to make sure that the previously selected recipients are remembered when navigating back and forth after selecting them for the first time. In the current development branch, if you select your recipients, hit continue, then go back, you will have to reselect your recipients. I think that is the only change to the service recipients.

Does that answer your questions?

samMcCarthyBJSS (approved):
Yeah those comments are fair!

PR #382, opened 2020-11-12
Feature/10328 bulk save bugfix

bjss-chris:
Story: [AB#10328](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/10328)
Task: [AB#10335](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/10335)

### DIFF
--- a/app/helpers/common/transformApiValidationResponse.js
+++ b/app/helpers/common/transformApiValidationResponse.js
@@ -1,0 +1,22 @@
+const transformPropertyName = propertyName => (propertyName.startsWith('[')
+  ? propertyName.substring(4)
+  : propertyName);
+
+export const transformApiValidationResponse = (validationResponse) => {
+  const distinctApiErrors = Object
+    .entries(validationResponse)
+    .reduce((acc, [key, value]) => {
+      const field = transformPropertyName(key);
+
+      if (acc[field] === undefined) {
+        acc[field] = new Set();
+      }
+
+      value.forEach(v => acc[field].add(v));
+      return acc;
+    }, {});
+
+  return Object
+    .entries(distinctApiErrors)
+    .flatMap(([key, value]) => [...value].map(v => ({ field: key, id: v })));
+};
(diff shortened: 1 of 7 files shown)


app/helpers/common/transformApiValidationResponse.js
```diff
@@ -0,0 +1,22 @@
+const transformPropertyName = propertyName => (propertyName.startsWith('[')
+  ? propertyName.substring(4)
+  : propertyName);
+
+export const transformApiValidationResponse = (validationResponse) => {
+  const distinctApiErrors = Object
```
samMcCarthyBJSS, line 6: can the lambdas be jojned to save making an object that's just going to be returned?

app/helpers/common/transformApiValidationResponse.test.js
```diff
@@ -0,0 +1,68 @@
+import { transformApiValidationResponse } from './transformApiValidationResponse';
+
+describe('transformApiValidationResponse', () => {
+  it('should transform a non-bulk validation error response', () => {
+    const field = 'DeliveryDate';
```
samMcCarthyBJSS, line 5: could this be a global const?

samMcCarthyBJSS (commented):
Left a couple of comments

app/helpers/common/transformApiValidationResponse.js
bjss-chris: Do you mean something like this?

```javascript
export const transformApiValidationResponse = validationResponse => Object.entries(Object
  .entries(validationResponse)
  .reduce((acc, [key, value]) => {
    const field = transformPropertyName(key);

    if (acc[field] === undefined) {
      acc[field] = new Set();
    }

    value.forEach(v => acc[field].add(v));
    return acc;
  }, {}))
  .flatMap(([key, value]) => [...value].map(v => ({ field: key, id: v })));
```
I think the existing version is more readable and I'm not sure we're really gaining much by joining the two expressions. My preference would be to stick with the more readable version.

samMcCarthyBJSS: yeah did mean that
although was going to suggest then swapping to a foreach or something for the entries bit
but yeah is more readable how you had it

samMcCarthyBJSS (approved):

PR #383, opened 2020-11-16
Only include service recipient when required

bjss-chris:
Story: [AB#10328](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/10328)
Task: [AB#10341](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/10341)

### DIFF
--- a/app/helpers/api/ordapi/postOrderItem.test.js
+++ b/app/helpers/api/ordapi/postOrderItem.test.js
@@ -96,6 +96,49 @@ describe('postOrderItem', () => {
       });
     });
 
+    it('should post correctly formatted data when there is no service recipient data', async () => {
+      const orderId = 'order1';
+
+      postData.mockResolvedValueOnce({ data: { orderId } });
+
+      const accessToken = 'access_token';
+      const catalogueSolutionId = 'some-solution-id';
+      const orderItemType = 'AssociatedService';
+
+      await postOrderItem({
+        orderId,
+        orderItemId: 'neworderitem',
+        orderItemType,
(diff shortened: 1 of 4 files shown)


app/helpers/api/ordapi/postOrderItem.test.js
```diff
@@ -96,6 +96,49 @@ describe('postOrderItem', () => {
...
+      const accessToken = 'access_token';
+      const catalogueSolutionId = 'some-solution-id';
+      const orderItemType = 'AssociatedService';
+
+      await postOrderItem({
+        orderId,
+        orderItemId: 'neworderitem',
+        orderItemType,
+        accessToken,
+        serviceRecipientId: null,
```
samMcCarthyBJSS, line 113: this should be undefined as well / we don't actually need to include them in the method call

samMcCarthyBJSS (commented):
just the one comment rest looks good

app/helpers/api/ordapi/postOrderItem.test.js
bjss-chris: This inclusion was actually intentional. I wanted to make sure the method worked as expected if one of the service recipient fields was `null` rather than `undefined`. I'm not sure if that should ever happen but I would've thought it is possible (even if unintended – i.e. developer error) so wanted to cover it.

samMcCarthyBJSS (approved):

PR #388, opened 2020-11-26
persist date bug

samMcCarthyBJSS:
Story [AB#10110](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/10110)
Task [AB#10454](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/10454)

### DIFF
--- a/app/helpers/routes/getCommencementDate.js
+++ b/app/helpers/routes/getCommencementDate.js
@@ -9,14 +9,14 @@ export const getCommencementDate = async ({
 }) => {
   const apiCall = async () => {
     const orgId = req.user.primaryOrganisationId;
-    return getCommencementDateFromApi({
+    const { commencementDate } = await getCommencementDateFromApi({
       orgId,
       accessToken,
       logger,
     });
+    return commencementDate;
   };
 
   const sessionData = { req, key: sessionKeys.plannedDeliveryDate };
-  const date = getFromSessionOrApi({ sessionData, sessionManager, apiCall });
-  return date.commencementDate ? date.commencementDate : date;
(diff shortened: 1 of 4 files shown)


app/helpers/routes/getCommencementDate.js
```diff
@@ -17,5 +17,6 @@ export const getCommencementDate = async ({
   };
 
   const sessionData = { req, key: sessionKeys.plannedDeliveryDate };
-  return getFromSessionOrApi({ sessionData, sessionManager, apiCall });
+  const date = getFromSessionOrApi({ sessionData, sessionManager, apiCall });
+  return date.commencementDate ? date.commencementDate : date;
```
bjss-chris, line 21: When will `date.commencementDate` be `undefined`?

samMcCarthyBJSS: if it gets it from session then it will only have the date as a string

bjss-chris: Ah, I see. Might it be worth tweaking `apiCall` to something like this instead? (Caveat: this code may not be 100% correct!)

```javascript
const apiCall = async () => {
    const orgId = req.user.primaryOrganisationId;
    const { commencementDate } = getCommencementDateFromApi({
      orgId,
      accessToken,
      logger,
    });

   return commencementDate;
};
```

What do you think?

samMcCarthyBJSS: yeah bit easier than having date.commencementdate to read

bjss-chris (approved):

PR #394, opened 2020-12-03
Feature/10509 session storage bug

samMcCarthyBJSS:
Story [AB#10509](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/10509)
Task [AB#10624](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/10624)

### DIFF
--- a/app/pages/sections/order-items/catalogue-solutions/routes.js
+++ b/app/pages/sections/order-items/catalogue-solutions/routes.js
@@ -113,6 +113,17 @@ export const catalogueSolutionsRoutes = (authProvider, addContext, sessionManage
 
       if (apiResponse.success) {
         logger.info('redirecting catalogue solutions main page');
+        sessionManager.saveToSession({ req, key: sessionKeys.solutions, value: undefined });
+        sessionManager.saveToSession({ req, key: sessionKeys.selectedItemId, value: undefined });
+        sessionManager.saveToSession({ req, key: sessionKeys.selectedItemName, value: undefined });
+        sessionManager.saveToSession({ req, key: sessionKeys.solutionPrices, value: undefined });
+        sessionManager.saveToSession({ req, key: sessionKeys.selectedPriceId, value: undefined });
+        sessionManager.saveToSession(
+          { req, key: sessionKeys.selectedRecipients, value: undefined },
+        );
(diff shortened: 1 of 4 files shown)


app/pages/sections/order-items/catalogue-solutions/routes.js
```diff
@@ -113,6 +113,17 @@ export const catalogueSolutionsRoutes = (authProvider, addContext, sessionManage
...
+        sessionManager.saveToSession({ req, key: sessionKeys.selectedItemId, value: undefined });
+        sessionManager.saveToSession({ req, key: sessionKeys.selectedItemName, value: undefined });
+        sessionManager.saveToSession({ req, key: sessionKeys.solutionPrices, value: undefined });
+        sessionManager.saveToSession({ req, key: sessionKeys.selectedPriceId, value: undefined });
+        sessionManager.saveToSession(
+          { req, key: sessionKeys.selectedRecipients, value: undefined },
+        );
+        sessionManager.saveToSession(
+          { req, key: sessionKeys.plannedDeliveryDate, value: undefined },
+        );
```
bjss-chris, line 126: Would `sessionManager.clearSession` be of any use here or are there some details that need to remain in session at this point (i.e. it's too destructive)?

bjss-chris (commented):
Just one question. Otherwise, looks good.

app/pages/sections/order-items/catalogue-solutions/routes.js
samMcCarthyBJSS: I did think about that but reckon we'll need them for other bits so probs better to not get rid of everything

bjss-chris: Do you think it's worth extracting this to a separate helper then? Could it be useful for additional and/or associated services?

samMcCarthyBJSS: Hm if it does get reused we should remove it but it's only used here currently
they won't have the solutions session key I don't think either

bjss-chris (approved):

PR #404, opened 2021-01-14
Only include service recipient when required

bjss-chris:
Activity: [AB#11185](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/11185)
Task: [AB#11195](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/11195)

### DIFF
--- a/app/helpers/api/ordapi/postOrderItem.js
+++ b/app/helpers/api/ordapi/postOrderItem.js
@@ -14,10 +14,10 @@ const formatPostData = ({
   formData,
 }) => ({
   ...selectedPrice,
-  serviceRecipient: {
+  serviceRecipient: serviceRecipientName || serviceRecipientId ? {
     name: serviceRecipientName,
     odsCode: serviceRecipientId,
-  },
+  } : undefined,
   catalogueItemId: itemId,
   catalogueItemName: itemName,
   catalogueItemType: orderItemType,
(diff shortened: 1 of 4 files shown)


app/helpers/api/ordapi/postOrderItem.js
```diff
@@ -14,10 +14,10 @@ const formatPostData = ({
   formData,
 }) => ({
   ...selectedPrice,
```
bjss-andrew, line 16: so serviceRecipient === serviceRecipientName if serviceRecipientName exists
but if it doesnt we set serviceRecipient to an object with a name propery

does that mean serviceRecipient will either be a string or an object (assuming serviceRecipientName is a string name and not an object)?

bjss-chris: `serviceRecipient` is always an object but we should only include it in the request to the API if we have the name and/or ID of the service recipient.

To be considered a valid request by the API, a service recipient must contain both a name and ID but a service recipient isn't always required. While we should never have only a name or an ID I thought it made sense to use an `or` condition rather than `and` so that the API validation is triggered if we ever encounter that situation, due to developer error say. Including a service recipient without a name or ID will always trigger the API validation but it will never be valid so shouldn't be included in this case.

Does that clarify things at all?

PhysicsJon (approved):

PR #408, opened 2021-01-19
Feature/pdf print font size

bjss-andrew:
override the table font only on printing

### DIFF
--- a/app/pages/summary/_summary.scss
+++ b/app/pages/summary/_summary.scss
@@ -1,0 +1,5 @@
+@media print {
+  .summary-page-print .nhsuk-u-font-size-14 {
+    font-size: 0.75rem !important;
+  }
+}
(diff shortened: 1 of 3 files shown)


app/pages/summary/_summary.scss
```diff
@@ -0,0 +1,5 @@
+@media print {
+  .summary-page-print .nhsuk-u-font-size-14 {
+    font-size: 0.75rem !important;
+  }
+}
```
bjss-chris, line 5: Just missing a trailing newline.

bjss-chris (changes requested):
Just an EOF to fix but otherwise LGTM.

app/pages/summary/_summary.scss
bjss-andrew: done

bjss-chris (approved):

PR #411, opened 2021-01-20
inline print script requires security policy to allow

bjss-andrew:
story #8739

### DIFF
--- a/app/pages/summary/routes.js
+++ b/app/pages/summary/routes.js
@@ -21,6 +21,7 @@ export const summaryRoutes = (authProvider, addContext) => {
     });
 
     if (print) {
+      res.set('Content-Security-Policy', 'default-src * \'unsafe-inline\' \'unsafe-eval\'; script-src * \'unsafe-inline\' \'unsafe-eval\';');
       return res.render('pages/summary/templatePrint.njk', addContext({ context, user: req.user }));
     }
 


app/pages/summary/routes.js
```diff
@@ -21,6 +21,7 @@ export const summaryRoutes = (authProvider, addContext) => {
     });
 
     if (print) {
+      res.set('Content-Security-Policy', 'default-src * \'unsafe-inline\' \'unsafe-eval\'; script-src * \'unsafe-inline\' \'unsafe-eval\';');
```
bjss-chris, line 24: I think this could be related to the updated `helmet` package as this has a content security policy  – see [here](https://github.com/helmetjs/helmet). Is it worth setting the helmet security policy? I guess the downside with setting the helmet policy is that it will apply everywhere in the order-form. What do you think?

I think we'll need to apply this to master too so it might be better to merge this to master and then we can merge master into development.

bjss-andrew: personally, I'd say to only relax the permission on the pages its needed, such as the print page, rather than globally in helmet, but happy to change if people would prefer a blanket change.

bjss-chris: I agree – it makes sense to only relax where necessary.

bjss-chris (approved):

PR #414, opened 2021-01-25
Fix bug with editing prices on bulk orders

bjss-andrew:
story #10751
task #11053


bjss-chris (commented):
Changes look fine but the checks have failed. Will approve after you fix.

PR #423, opened 2021-02-05
Increment bcc npm package

NickNewlands:
Fixed lint:css issue

task: [AB#11862](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/11862)
story: [AB#8971](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/8971)

The amend to the CSS linting rules is because the nhsuk-frontend library uses !important:
https://github.com/nhsuk/nhsuk-frontend/blob/master/packages/core/utilities/_typography.scss

```
// Utility classes are allowed to use !important;
// so we disable the sass-lint for that rule
// sass-lint:disable no-important
```


.sass-lint.yml
```diff
@@ -35,7 +35,7 @@ rules:
   no-empty-rulesets: 2
   no-extends: 0
   no-ids: 2
-  no-important: 2
+  no-important: 0
```
bjss-chris, line 38: Why do we need to change this linting rule?

bjss-chris (changes requested):
The tests are currently failing (in the pipeline). Can you review and fix?

.sass-lint.yml
NickNewlands: Because we need to use !important to override an !important coming from the base styling and this rule would not let us.

bjss-andrew: the print file has to override none print styles, but Id prob have a scss lint ignore comment in the print.scss with an explanation maybe

NickNewlands: It's a broad problem with using nhsuk-frontend, they are using !important so we have to to override:
https://github.com/nhsuk/nhsuk-frontend/blob/master/packages/core/utilities/_typography.scss

`// Utility classes are allowed to use !important;
// so we disable the sass-lint for that rule
// sass-lint:disable no-important`

bjss-chris: Ah, OK. That makes sense now. Thanks for the reference.

Might be worth adding a comment just stating that the no-important rule is disabled due to the NHS library though.

bjss-chris:
Shouldn't we also update `package-lock.json`? Recommend using `npm` to update the package reference so that `package-lock.json` is also updated.

bjss-chris (changes requested):
Just needs a comment adding to the `.sass-lint.yml` file (see conversation) and `package-lock.json` updating. Otherwise, the changes look good.

bjss-chris (approved):

PR #425, opened 2021-02-05
Price displays to 4 decimal places

NickNewlands:
story: [AB#11143](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/11143)
task: [AB#11722](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/11722)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "order-form",
-  "version": "1.0.1",
+  "version": "1.1.0",
```
NickNewlands, line 3: I am not sure if this should be a patch or minor.

app/helpers/common/priceFormatter.js
```diff
@@ -1,16 +1,22 @@
...
+    minimumFractionDigits: 2, maximumFractionDigits: 4,
   });
 
-  const truncatedTo2dp = formattedPrice.substring(0, formattedPrice.indexOf('.') + 3);
-
-  return truncatedTo2dp;
+  return formattedPrice;
 };
 
 export const formatDecimal = (priceValue) => {
```
bjss-andrew, line 9: this function returns either a number or a string. id prob say it should be consistent, but not major if its the only change

package.json
bjss-chris: I'd probably treat this as a patch. However, we use git versioning for the actual applications so this is less important here but it doesn't do any harm to update the `package.json` version as well.

Git versioning should automatically apply a patch version although you can also explicitly state that the patch version should be increased. There a couple of ways to do this but I tend to find the simplest is to include `+semver:patch` in your commit message body – this is another thing that I need to document.

NickNewlands: I have updated it to a patch, looks like a new package-lock was generated, I expect that is fine.

app/helpers/common/priceFormatter.js
NickNewlands: I think regarding the UI, it's ok to always return a string as we should not be doing any maths anyway and just displaying.

app/helpers/common/priceFormatter.js
```diff
@@ -1,16 +1,22 @@
...
+  if (priceStringArr.length === 1) return priceStringArr[0];
+
+  const fractionalLength = priceStringArr[1].length;
+  if (fractionalLength < 5) {
+    return parseFloat(priceValue).toFixed(fractionalLength < 2 ? 2 : fractionalLength);
   }
-  return priceValue;
+
+  return parseFloat(priceValue).toFixed(4);
 };
```
bjss-chris, line 22: I think we ought to do a bit more refactoring here. I actually think we don't need two functions. There seems to be two main differences between the original versions of `formatPrice` and `formatDecimal`:

1. `formatPrice` returns a string, whereas `formatDecimal` returns a number
2. `formatPrice` always formats to 2 d.p. but `formatDecimal` only formats numbers with a fractional part.

However, in the majority of places `formatDecimal` appears to be used to format numbers for display to users (where returning a string is obviously fine) although there are some places where it appears to be used to parse user input. Other than that the names are confusing. It isn't clear which one a developer should use. Therefore I suggest we should refactor `formatPrice` to this:


```javascript
export const formatPrice = ({ minimumFractionDigits = 2, value }) =>
    parseFloat(value).toLocaleString(undefined, { minimumFractionDigits, maximumFractionDigits: 4 });
```

We then have two options for `formatDecimal` – we refactor it to this:

```javascript
export const formatDecimal = (priceValue) => formatPrice({ minimumFractionDigits: 0, value: priceValue });
```

or we remove it completely.

We will, of course, have to update the usages of `formatPrice` to `formatPrice({ value: whatever })` and if we remove `formatDecimal` we will also to have update its existing usages to `formatPrice({ minimumFractionDigits: 0, value: whatever })`.

I'm in two minds about whether to keep `formatDecimal`, although keeping it means fewer usage updates and could be said to keep things DRY (by reducing `formatPrice({ minimumFractionDigits: 0, value: whatever })` to `formatDecimal(value)`). However, if we do keep it then I think we definitely need to improve naming. Ideally indicate in some way that one always includes decimal places but the other doesn't. And this is why I'm in two minds because I suspect the naming could be awkward and updating `formatDecimal` uses to `formatPrice({ minimumFractionDigits: 0, value: whatever })` makes it absolutely clear that we don't require decimal places in that instance.

Where `formatDecimal` is currently used to parse user input I think we should just be able to replace it with `parseFloat`.

Sorry about the TL;DR. I'll be happy to clarify anything on Monday.

app/pages/summary/complete/ui-tests/withSavedData.ui.test.js
```diff
@@ -343,10 +343,10 @@ test('should render the recurring cost item details in the table', async (t) =>
     .expect(await extractInnerText(recurringCostRow1.find('td').nth(1))).eql('C000001-01-A10001-2')
     .expect(await extractInnerText(recurringCostRow1.find('td').nth(2))).eql('Some catalogue name')
     .expect(await extractInnerText(recurringCostRow1.find('td').nth(3))).eql('SI1-A10001')
-    .expect(await extractInnerText(recurringCostRow1.find('td').nth(4))).eql('500.26 per license')
+    .expect(await extractInnerText(recurringCostRow1.find('td').nth(4))).eql('500.261 per license')
     .expect(await extractInnerText(recurringCostRow1.find('td').nth(5))).eql('12')
     .expect(await extractInnerText(recurringCostRow1.find('td').nth(6))).eql('6 August 2020')
-    .expect(await extractInnerText(recurringCostRow1.find('td').nth(7))).eql('6,003.13')
+    .expect(await extractInnerText(recurringCostRow1.find('td').nth(7))).eql('6,003.132')
```
bjss-chris, line 349: I think it might be worth including a price with 4 d.p. in one or more of the UI tests. (Either this one and/or the UI tests for an incomplete order.)

app/pages/summary/contextCreator.test.js
```diff
@@ -625,7 +625,7 @@ describe('order summary contextCreator', () => {
                 dataTestId: 'total-monthly-cost-label',
               },
               {
-                data: '191.69',
+                data: '191.691',
```
bjss-chris, line 628: Worth using 4 d.p.?

app/pages/summary/incomplete/ui-tests/withSavedData.ui.test.js
```diff
@@ -343,10 +343,10 @@ test('should render the recurring cost item details in the table', async (t) =>
     .expect(await extractInnerText(recurringCostRow1.find('td').nth(1))).eql('C000001-01-A10001-2')
     .expect(await extractInnerText(recurringCostRow1.find('td').nth(2))).eql('Some catalogue name')
     .expect(await extractInnerText(recurringCostRow1.find('td').nth(3))).eql('SI1-A10001')
-    .expect(await extractInnerText(recurringCostRow1.find('td').nth(4))).eql('500.26 per license')
+    .expect(await extractInnerText(recurringCostRow1.find('td').nth(4))).eql('500.261 per license')
     .expect(await extractInnerText(recurringCostRow1.find('td').nth(5))).eql('12')
     .expect(await extractInnerText(recurringCostRow1.find('td').nth(6))).eql('6 August 2020')
-    .expect(await extractInnerText(recurringCostRow1.find('td').nth(7))).eql('6,003.13')
+    .expect(await extractInnerText(recurringCostRow1.find('td').nth(7))).eql('6,003.132')
```
bjss-chris, line 349: I think it might be worth including a price with 4 d.p. in one or more of the UI tests. (Either this one and/or the UI tests for a complete order.)

bjss-chris (changes requested):

app/helpers/common/priceFormatter.js
bjss-chris: After thinking about this a bit more I think something like this might work quite nicely:

```javascript
export const formatNumber = ({ minimumFractionDigits = 0, maximumFractionDigits = 4, value }) =>
    parseFloat(value).toLocaleString(undefined, { minimumFractionDigits, maximumFractionDigits });

export const formatPrice = (value) => formatNumber({ minimumFractionDigits: 2, value });
```

app/pages/summary/complete/ui-tests/withSavedData.ui.test.js
bjss-andrew: id include one with 5/6 decimal places too

app/helpers/common/priceFormatter.js
```diff
@@ -1,16 +1,8 @@
-export const formatPrice = (priceValue) => {
-  const formattedPrice = priceValue.toLocaleString(undefined, {
-    minimumFractionDigits: 3, maximumFractionDigits: 3,
-  });
+export const formatNumber = ({ value, minimumFractionDigits = 0, maximumFractionDigits = 4 }) => (
+  parseFloat(value)
+    .toLocaleString(
+      undefined, { minimumFractionDigits, maximumFractionDigits },
+    )
+);
```
NickNewlands, line 7: I don't love this format but I really had to try hard to find something eslint agreed with.

It is currently setup to not like line-breaks for arrow functions but insists on 100 character max line length so wrapping in brackets is necessary.

app/pages/sections/order-items/catalogue-solutions/select/price/contextCreator.test.js
```diff
@@ -75,7 +75,7 @@ const returnedPriceArray = [{
     value: 2,
     checked: true,
   }, {
-    html: '<div>1 - 10 consultations £700 per consultation per month</div><div>11+ consultations £400 per consultation per month</div>',
+    html: '<div>1 - 10 consultations £700.00 per consultation per month</div><div>11+ consultations £400.00 per consultation per month</div>',
```
NickNewlands, line 78: I need to ask Lorraine if this is a problem, I believe this is the only place a price is displayed as a whole number.

app/pages/summary/incomplete/ui-tests/withSavedData.ui.test.js
```diff
@@ -343,10 +343,10 @@ test('should render the recurring cost item details in the table', async (t) =>
     .expect(await extractInnerText(recurringCostRow1.find('td').nth(1))).eql('C000001-01-A10001-2')
     .expect(await extractInnerText(recurringCostRow1.find('td').nth(2))).eql('Some catalogue name')
     .expect(await extractInnerText(recurringCostRow1.find('td').nth(3))).eql('SI1-A10001')
-    .expect(await extractInnerText(recurringCostRow1.find('td').nth(4))).eql('500.26 per license')
+    .expect(await extractInnerText(recurringCostRow1.find('td').nth(4))).eql('500.261 per license')
     .expect(await extractInnerText(recurringCostRow1.find('td').nth(5))).eql('12')
     .expect(await extractInnerText(recurringCostRow1.find('td').nth(6))).eql('6 August 2020')
-    .expect(await extractInnerText(recurringCostRow1.find('td').nth(7))).eql('6,003.13')
+    .expect(await extractInnerText(recurringCostRow1.find('td').nth(7))).eql('6,003.132245')
```
NickNewlands, line 350: This test seems incorrect, it must not be running.

PR #475, opened 2021-03-01
Move order summary button

NickNewlands:
From right to left for accessibility

Task: [AB#12882](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/12882)
Story: [AB#12166](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/12166)

getSummary

### DIFF
--- a/app/pages/summary/template.njk
+++ b/app/pages/summary/template.njk
@@ -120,17 +120,17 @@
         </div>
       </div>
 
-      <div class="bc-u-float-right nhsuk-grid-column-one-third">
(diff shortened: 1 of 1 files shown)


NickNewlands:
I really don't like this but I suppose it is more accessible. There is now a button on the right and left so you can't miss it but I found that zooming the browser would put you into mobile/tablet mode so I don't think this change helps much.

bjss-chris (approved):

PR #487, opened 2021-03-08
update post request to a put with new endpoint url

bjss-andrew:
api contract has now changed so that rather than the /bulk url fragment its takes the item id and changes from a http post to a http put

### DIFF
--- a/app/helpers/api/ordapi/postOrderItemBulk.js
+++ b/app/helpers/api/ordapi/postOrderItemBulk.js
@@ -1,4 +1,4 @@
-import { postData } from 'buying-catalogue-library';
+import { putData } from 'buying-catalogue-library';
 import { logger } from '../../../logger';
 import { extractDate } from '../../controllers/extractDate';
 import { orderApiUrl } from '../../../config';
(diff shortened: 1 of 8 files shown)


app/helpers/api/ordapi/postOrderItemBulk.js
```diff
@@ -27,7 +27,7 @@ const formatPostData = ({
   price: parseFloat(formData.price),
 }));
 
-const getPostOrderItemEndpoint = (orderId) => `${orderApiUrl}/api/v1/orders/${orderId}/order-items/batch`;
+const getOrderItemEndpoint = (orderId, itemId) => `${orderApiUrl}/api/v1/orders/${orderId}/order-items/${itemId}`;
```
bjss-chris, line 30: `itemId` is actually the catalogue item ID now (e.g. 10001-001), so it might be better to rename this to `catalogueItemId` just to avoid any confusion around people thinking it is an arbitrary order item ID (e.g. 1, 2, 3 etc.).

app/helpers/api/ordapi/postOrderItemBulk.test.js
```diff
@@ -77,9 +77,9 @@ describe('postOrderItemBulk', () => {
...
 
-      expect(postData.mock.calls.length).toEqual(1);
-      expect(postData).toHaveBeenCalledWith({
-        endpoint: `${orderApiUrl}/api/v1/orders/order1/order-items/batch`,
+      expect(putData.mock.calls.length).toEqual(1);
+      expect(putData).toHaveBeenCalledWith({
+        endpoint: `${orderApiUrl}/api/v1/orders/order1/order-items/item-1`,
         body: [{
           ...selectedPrice,
           serviceRecipient,
```
bjss-chris, line 85: The new contract no longer expects an array so it looks like further changes are required.

The request body (in pseudo-form) should be like this:


```json
{
    "catalogueItemName": "value",
    "serviceRecipients": [
        {
            "odsCode": "ABC1",
            "quantity": 1,
            "otherProperties..."
        }
    ]
    "otherProperties..."
}
```

The updated contract is now on the [wiki](https://dev.azure.com/BuyingCatalog/Buying%20Catalogue/_wiki/wikis/Wiki/183/BE-ORDAPI-bulk?anchor=put-%2Fapi%2Fv1%2Forders%2F%7Bcalloffid%7D%2Forder-items%2F%7Bcatalogueitemid%7D).

bjss-chris (changes requested):
I think additional changes are required (see my comment in the tests).

Updated contract is now on the [wiki](https://dev.azure.com/BuyingCatalog/Buying%20Catalogue/_wiki/wikis/Wiki/183/BE-ORDAPI-bulk?anchor=put-%2Fapi%2Fv1%2Forders%2F%7Bcalloffid%7D%2Forder-items%2F%7Bcatalogueitemid%7D).

Let me know if you've any questions.

bjss-andrew:
as an fyi, our code has the http verb in the file names and method names, what should be a little change from a  POST to a PUT is a big refactor, id like to add that a tech debt as renaming anything wont add value but delay the bug

app/helpers/api/ordapi/postOrderItemBulk.js
```diff
@@ -40,20 +39,23 @@ export const postOrderItemBulk = async ({
   recipients,
   formData,
 }) => {
-  const endpoint = getPostOrderItemEndpoint(orderId);
+
+  const endpoint = `${orderApiUrl}/api/v1/orders/${orderId}/order-items/${itemId}`;
```
bjss-chris, line 43: It would be better if `itemId` was `catalogueItemId` but we can create a tech debt task for that if necessary as it would be good to get this fix merged.

bjss-chris (commented):
Just one comment but it's possibly for one tech debt. There's a test failing in the pipeline that we'll need to fix but otherwise looks good.

bjss-chris (approved):

PR #491, opened 2021-03-09
Update nhsuk-frontend npm package

NickNewlands:
Story: [AB#11214](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/11214)
Ticket: [AB#13088](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/13088)

### DIFF
--- a/app/pages/sections/commencement-date/template.test.js
+++ b/app/pages/sections/commencement-date/template.test.js
@@ -98,7 +98,7 @@ describe('commencement-date page', () => {
 
     it('should render additionalAdvice', componentTester(setup, (harness) => {
       harness.request(context, ($) => {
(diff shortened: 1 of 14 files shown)


NickNewlands:
Seems like there is not a big impact in upgrading the package but the 'type' property has been removed off of the input.

bjss-chris:
> Seems like there is not a big impact in upgrading the package but the 'type' property has been removed off of the input.

Yep, looks like they removed that and set `inputmode="numeric"` instead – see the [change log](https://github.com/nhsuk/nhsuk-frontend/blob/master/CHANGELOG.md#410---21-january-2021) for details (second bullet point in new features list).

bjss-chris (approved):

PR #494, opened 2021-03-10
Feature/12166 accessibility audit

NickNewlands:


bjss-chris (dismissed):
I must have missed them before but there are some prefixes in some of the tests that we should remove, e.g. `tFoot` would be better as `foot`, `footer` or even just `table` as it was before. However, we can make those tweaks under another task.

bjss-andrew (dismissed):

bjss-chris (approved):

bjss-chris (dismissed):

bjss-chris:
Looks like the tests have timed-out in the pipeline. Does it need your `process.exit` fix @NickNewlands?

KeithLyonMastek (approved):

bjss-chris (approved):

PR #510, opened 2021-03-18
price validation bug fix

nehajain008:
Story [AB#12945](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/12945)
Task [AB#13189](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/13189)

### DIFF
--- a/app/helpers/common/priceFormatter.js
+++ b/app/helpers/common/priceFormatter.js
@@ -9,10 +9,13 @@ export const formatNumber = ({
 
 export const formatPrice = ({ value, minimumFractionDigits = 2, maximumFractionDigits = 4 }) => {
   if (minimumFractionDigits === 0) {
-    if (Number.isInteger(Number.parseFloat(value))) {
-      return formatNumber({ value, maximumFractionDigits });
+    if (!Number.isNaN(Number(value)) && !Number.isNaN(parseFloat(value))) {
+      if (Number.isInteger(Number.parseFloat(value))) {
+        return formatNumber({ value, maximumFractionDigits });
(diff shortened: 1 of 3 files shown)


bjss-chris:
@nehajain008, are there any additional test cases we can add to cover this scenario?

app/helpers/common/priceFormatter.js
```diff
@@ -9,10 +9,13 @@ export const formatNumber = ({
 
 export const formatPrice = ({ value, minimumFractionDigits = 2, maximumFractionDigits = 4 }) => {
   if (minimumFractionDigits === 0) {
-    if (Number.isInteger(Number.parseFloat(value))) {
-      return formatNumber({ value, maximumFractionDigits });
+    if (!Number.isNaN(Number(value)) && !Number.isNaN(parseFloat(value))) {
```
bjss-andrew, line 12: it might work, but the logic here is not easy to follow, would benefit from a refactor

nehajain008: done few changes as a part of refactoring

nehajain008:
> 
> 
> @nehajain008, are there any additional test cases we can add to cover this scenario?

added testcase for new scenario

bjss-chris (approved):
I think we might one to revisit the price formatter again at some point as tech debt but I this is definitely an improvement over the previous version.

@bjss-andrew, what do you think?

PR #512, opened 2021-03-19
Display Additional Services in alphabetical order

kiran-mastek:
Sort additional services

Story: [AB#9227](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/9227)
Task: [AB#13222](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/13222)

### DIFF
--- a/app/helpers/api/bapi/getAdditionalServices.js
+++ b/app/helpers/api/bapi/getAdditionalServices.js
@@ -1,6 +1,7 @@
 import { getData } from 'buying-catalogue-library';
 import { logger } from '../../../logger';
 import { solutionsApiUrl } from '../../../config';
+import { sortItems } from '../../common/sortItems';
 
 const getAdditionalServicesEndpoint = (addedCatalogueSolutions) => {
   const queryString = `solutionIds=${addedCatalogueSolutions.join('&solutionIds=')}`;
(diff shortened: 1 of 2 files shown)


app/helpers/api/bapi/getAdditionalServices.js
```diff
@@ -14,6 +15,8 @@ export const getAdditionalServices = async ({ addedCatalogueSolutions, accessTok
     return [];
   }
 
+  sortItems(additionalServices, 'name');
```
KeithLyonMastek, line 18: We seem to have 2 different ways of using sortItems.  getOrder.js assigns the output of sortItems to a variable, where as getCatalogueItems.js updates the variable inside the function, like what has been done here.

I believe it is best to assign the output of the sortItems function to a variable, rather than change the value inside a function, as this can cause confusion.

Would it be worth looking at changing the use of sortItems to be consistent between all files ?

kiran-mastek: done

bjss-chris (approved):
Looks good to me. You happy @KeithLyonMastek?

We can add a tech debt ticket to update `getCatalogueItems` and any other places that we are currently doing an in-place sort to follow the same approach as here.

PR #513, opened 2021-03-19
save session on dashboard and redirect to edit page

nehajain008:
Story [AB#9324](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/9324)
Task [AB#13221](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/13221)
Task [AB#13238](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/13238)

### DIFF
--- a/app/pages/sections/order-items/catalogue-solutions/select/routes.js
+++ b/app/pages/sections/order-items/catalogue-solutions/select/routes.js
@@ -80,6 +80,14 @@ export const catalogueSolutionsSelectRoutes = (authProvider, addContext, session
         catalogueItemsKey: 'solutions',
       });
 
+      const orderItems = sessionManager.getFromSession({ req, key: sessionKeys.orderItems });
+      const alreadySelectedItem = orderItems
+        .filter((orderItem) => orderItem.catalogueItemId === selectedItem.catalogueItemId);
+      if (alreadySelectedItem.length > 0 && alreadySelectedItem[0].catalogueItemId) {
+        return res.redirect(
+          `${config.baseUrl}/organisation/${orderId}/catalogue-solutions/${alreadySelectedItem[0].catalogueItemId}`,
+        );
+      }
       sessionManager.saveToSession({
         req, key: sessionKeys.selectedItemId, value: selectedItemId,
       });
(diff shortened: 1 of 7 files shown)


app/pages/sections/order-items/catalogue-solutions/select/routes.js
```diff
@@ -80,6 +80,14 @@ export const catalogueSolutionsSelectRoutes = (authProvider, addContext, session
...
       });
 
+      const orderItems = sessionManager.getFromSession({ req, key: sessionKeys.orderItems });
+      const alreadySelectedItem = orderItems
+        .filter((orderItem) => orderItem.catalogueItemId === selectedItem.catalogueItemId);
+      if (alreadySelectedItem.length > 0 && alreadySelectedItem[0].catalogueItemId) {
+        return res.redirect(
+          `${config.baseUrl}/organisation/${orderId}/catalogue-solutions/${alreadySelectedItem[0].catalogueItemId}`,
+        );
+      }
```
bjss-chris, line 90: What happens if we don't have the order items in session, such as if the user has created an order without completing it and comes back the next day? I think the catalogue solutions selection page will still be displayed in that case – is that correct? 

Therefore, we'll need to call the API (to get a list of the order items) if we don't have any order items in session won't we? We have a `getFromSessionOrApi` helper function, you can see some examples of its use in `helpers/routes/` (such as `getOrderDescription`). Could we use something similar here?

bjss-chris (changes requested):

app/pages/sections/order-items/catalogue-solutions/select/routes.js
nehajain008: thanks Chris , will check that scenario and getFromSessionOrApi helper

bjss-chris: Thanks for helping resolve my misunderstanding @nehajain008!

bjss-chris (approved):

PR #527, opened 2021-03-23
change get order item endpoint

bjss-andrew:


bjss-chris (commented):
Changes look fine but we've got some failing tests.

PR #535, opened 2021-03-24
Feature/13000 get order item

bjss-andrew:


bjss-chris:
@bjss-andrew can you review the code and confirm you're happy?

@PhysicsJon, can you review this? Both Andrew and I have worked on it so I'd prefer not to approve it myself. I'm happy with it. If @bjss-andrew is also happy with it then I suggest you just double-check it so we've had someone review it who hasn't contributed to the change.

PhysicsJon (approved):
Approved for now, is there a plan for all the commented out code in the tests?

bjss-andrew:
yes it will go back in, there is some other commented out code with a tech debt ticket to re-implement

PR #542, opened 2021-03-25
Add feature flags section to helm values

bjss-chris:
Story: [AB#9257](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/9257)
Task: [AB#13367](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/13367)

### DIFF
--- a/charts/of/values.yaml
+++ b/charts/of/values.yaml
@@ -10,7 +10,7 @@ image:
   repository: gpitfuturesdevacr.azurecr.io/nhsd/buying-catalogue/order-form
   pullPolicy: Always
 
-imagePullSecrets: 
+imagePullSecrets:
 nameOverride: "of"
 fullnameOverride: "gpitfutures-bc-of"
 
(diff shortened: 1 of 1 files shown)


PhysicsJon (approved):

bjss-andrew (approved):

bjss-chris:
@PhysicsJon, @bjss-andrew I'm going to hang fire until @gavinwakeley1 has tested this in a dynamic environment before merging.

We're not demoing additional services today so the feature flag defaulting to true and therefore breaking additional services shouldn't cause an immediate problem!

bjss-chris:
> @PhysicsJon, @bjss-andrew I'm going to hang fire until @gavinwakeley1 has tested this in a dynamic environment before merging.
> 
> We're not demoing additional services today so the feature flag defaulting to true and therefore breaking additional services shouldn't cause an immediate problem!

Actually looks like I will need to merge it, as the additional services changes references this section so the local cluster won't start up without it.

PR #553, opened 2021-03-29
use a smaller build base to reduce docker image size

bjss-andrew:
reduces the docker image size from 600mb to 180mb

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:14-slim as builder
+FROM node:14-slim
 
 # Create app directory
 WORKDIR /usr/src
@@ -17,7 +17,7 @@ RUN npm run build:docker
 # Install prod dependencies inside the dist directory
 RUN cp package.json dist && cd dist && npm install --only=prod
 
-FROM builder as production
+FROM node:14-slim
 
 COPY . .
 


PhysicsJon (approved):

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:14-slim as builder
+FROM node:14-slim
```
bjss-chris, line 1: Will this work using `node:14-alpine`?

Dockerfile
```diff
@@ -17,7 +17,7 @@ RUN npm run build:docker
 # Install prod dependencies inside the dist directory
 RUN cp package.json dist && cd dist && npm install --only=prod
 
-FROM builder as production
+FROM node:14-slim
```
bjss-chris, line 20: As above, will this work using `node:14-alpine`?

Dockerfile
```diff
@@ -17,7 +17,7 @@ RUN npm run build:docker
...
 
 COPY . .
```
bjss-chris, line 22: I was thinking about this yesterday, won't we need an alias for the images?

Shouldn't this `COPY` command include a `--from=build`, assuming the alias for the first image is `build`?

Without that won't this command just be copying files from the host?

PR #613, opened 2021-04-21
add showProxy and SHOW_PROXY flag

KeithLyonMastek:
Task [AB#13822](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/13822)

### DIFF
--- a/app/config.js
+++ b/app/config.js
@@ -72,4 +72,7 @@ module.exports = {
 
   // The password to connect to redis
   redisPass: process.env.REDIS_PASS,
+
+  // content and user journey if they can order for multiple organisations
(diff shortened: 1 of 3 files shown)


bjss-chris (changes requested):
Changes look fine. However, as this is the first order-form feature flag we'll need to add a feature flags section to the `charts/values.yaml`, so that it looks like this:

```yaml
feedbackLinkUrl:

featureFlags:
  showProxy: "true"

serviceDependencies:
```

I've suggested that location in the file so that it is similar to the other apps, e.g. public browse. Line 78 I think.

bjss-chris (approved):

PR #615, opened 2021-04-21
show proxy "Order for" text if flag set

KeithLyonMastek:
Task [AB#13821](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/13821)

### DIFF
--- a/app/config.js
+++ b/app/config.js
@@ -74,5 +74,5 @@ module.exports = {
   redisPass: process.env.REDIS_PASS,
 
   // content and user journey if they can order for multiple organisations
-  showProxy: process.env.SHOW_PROXY || 'true',
+  showProxy: process.env.SHOW_PROXY || 'false',
 };
--- a/app/config.js
+++ b/app/config.js
@@ -74,5 +74,5 @@ module.exports = {
   redisPass: process.env.REDIS_PASS,
 
   // content and user journey if they can order for multiple organisations
-  showProxy: process.env.SHOW_PROXY || 'false',
+  showProxy: process.env.SHOW_PROXY || 'true',
 };
--- a/app/pages/dashboard/contextCreator.js
+++ b/app/pages/dashboard/contextCreator.js
@@ -98,4 +98,5 @@ export const getContext = ({ orgName, completedOrders = [], incompletedOrders = 
   incompleteOrders: generateIncompletedOrdersTable(
     incompletedOrders, manifest.incompleteOrdersTable,
   ),
+  secondaryOrganisation: '',
 });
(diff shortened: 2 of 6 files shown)


app/config.js
```diff
@@ -74,5 +74,5 @@ module.exports = {
   redisPass: process.env.REDIS_PASS,
 
   // content and user journey if they can order for multiple organisations
-  showProxy: process.env.SHOW_PROXY || 'true',
+  showProxy: process.env.SHOW_PROXY || 'false',
```
bjss-chris, line 77: I think it might be best to leave this to default to true, unless you think it could cause any problems?

bjss-chris (commented):
Looks good to me – just one suggestion/question on the change of the default flag value.

app/config.js
KeithLyonMastek: The flag should only become true if the logged in user is a proxy buyer.  By default, they wont be

bjss-chris: Not sure I follow – the user's credentials won't have any effect on the flag. However, the story doesn't make it clear what the expected behaviour is for a non-proxy user, e.g. whether the dashboard should appear exactly as it does now, which would make sense but best not to assume – currently waiting for a response from Emily.

If the above is true then we'll need to add an extra condition to the template to make sure the proxy-version of the dashboard is only rendered for proxy users. Once the API changes have been completed the user object will have a `secondaryOrganisation` claim if they are a proxy buyer (`req.user.secondaryOrganisation`).

In dev, I think it will be best if we keep the default value of the flag as true though.

KeithLyonMastek: The ticket's requirements have been met and by having it as false ensures no scope creep

KeithLyonMastek: so the logic should be that is user credentials says your a proxy, show the proxy information.  

And the flag should just be "if flag is true, and if usercredentials.isProxy" then show?

bjss-chris: The flag should just be true or false (set by environment variable) or a default value if no environment variable present, which can be true as we'll explicitly set an environment variable in the non-dev environments.

Therefore, the condition that determines whether the proxy information is shown should be 'flag is true and user is proxy', e.g. something akin to `showProxy && req.user.secondaryOrganisation`.

KeithLyonMastek: I have implemented what I mentioned above, the if flag is true, and if usercredentials.isProxy", by creating a dummy variable (as the API has not been finished yet with the actual variable) to hold the 'usercredentials.isProxy' value

bjss-chris (approved):

PR #618, opened 2021-04-26
Feature/13824 add select organization page and get route

KeithLyonMastek:
Task [AB#13824](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/13824)

### DIFF
--- a/app/helpers/api/oapi/getOrganisation.js
+++ b/app/helpers/api/oapi/getOrganisation.js
@@ -12,10 +12,3 @@ export const getOrganisation = async ({ orgId, accessToken }) => {
   logger.info(`Organisation with id: ${orgId} found in OAPI`);
   return organisationData;
 };
-
-export const getRelatedOrganisations = async ({ accessToken, orgId }) => {
-  const orgDataEndpoint = `${getOrganisationEndpoint({ orgId })}/related-organisations`;
-  const organisationData = await getData({ endpoint: orgDataEndpoint, accessToken, logger });
-  logger.info('Get Related Organisations found in OAPI');
-  return organisationData;
-};
--- a/app/helpers/api/oapi/getOrganisation.js
+++ b/app/helpers/api/oapi/getOrganisation.js
@@ -12,3 +12,10 @@ export const getOrganisation = async ({ orgId, accessToken }) => {
   logger.info(`Organisation with id: ${orgId} found in OAPI`);
   return organisationData;
 };
+
+export const getRelatedOrganisations = async ({ accessToken, orgId }) => {
+  const orgDataEndpoint = `${getOrganisationEndpoint({ orgId })}/related-organisations`;
+  const organisationData = await getData({ endpoint: orgDataEndpoint, accessToken, logger });
+  logger.info('Get Related Organisations found in OAPI');
+  return organisationData;
+};
--- a/app/pages/dashboard/controller.test.js
+++ b/app/pages/dashboard/controller.test.js
@@ -7,6 +7,7 @@ jest.mock('./contextCreator', () => ({
   getContext: jest.fn(),
 }));
 jest.mock('../select/controller');
+jest.mock('../../helpers/controllers/getIsUserProxy');
 
 describe('dashboard controller', () => {
   describe('getDashboardContext', () => {
@@ -32,7 +33,9 @@ describe('dashboard controller', () => {
       await getDashboardContext({ orgId: 'org-id', orgName: 'org1', accessToken: 'access_token' });
 
       expect(contextCreator.getContext.mock.calls.length).toEqual(1);
-      expect(contextCreator.getContext).toHaveBeenCalledWith({ orgName: 'org1', completedOrders: [], incompletedOrders: [] });
+      expect(contextCreator.getContext).toHaveBeenCalledWith({
+        orgName: 'org1', completedOrders: [], incompletedOrders: [], userIsProxy: undefined,
+      });
     });
   });
 });
--- a/app/pages/dashboard/controller.test.js
+++ b/app/pages/dashboard/controller.test.js
@@ -1,6 +1,7 @@
 import { getDashboardContext } from './controller';
 import * as contextCreator from './contextCreator';
 import { getOrders } from '../../helpers/api/ordapi/getOrders';
+import { getIsUserProxy } from '../../helpers/controllers/getIsUserProxy';
 
 jest.mock('../../helpers/api/ordapi/getOrders');
 jest.mock('./contextCreator', () => ({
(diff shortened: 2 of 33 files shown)


app/pages/dashboard/controller.test.js
```diff
@@ -6,6 +6,7 @@ jest.mock('../../helpers/api/ordapi/getOrders');
 jest.mock('./contextCreator', () => ({
   getContext: jest.fn(),
 }));
+jest.mock('../select/controller');
```
bjss-chris, line 9: Can we update the the test `'should call getContext with the correct params when orders data is returned by getData'` to include the `userIsProxy` change?

app/pages/select/controller.js
```diff
@@ -0,0 +1,35 @@
...
+
+export const getSelectContext = async ({ accessToken, orgId, orgName }) => {
+  const organisationsList = await getProxyOrganisations({ accessToken, orgId });
+
+  let radioList;
+
+  if (organisationsList) {
+    radioList = organisationsList.map((org) => ({ value: org.organisationId, text: org.name }));
+    radioList.sort((a, b) => a.text.localeCompare(b.text));
+  }
```
bjss-chris, line 16: I think this might sit nicely in a separate function, e.g. `formatOrganisationList` or `transformOrganisationList` or something similar.

app/pages/select/controller.js
```diff
@@ -0,0 +1,35 @@
...
+export const getIsUserProxy = async ({ accessToken, orgId }) => {
+  if (!accessToken) {
+    return false;
+  }
+
+  const proxyOrganisations = await getProxyOrganisations({ accessToken, orgId });
+
+  const userIsProxy = proxyOrganisations.length > 0;
+  return userIsProxy;
+};
```
bjss-chris, line 35: I'm not sure this belongs in a controller. I think it might be better being moved out to a helper.

Also, why make an extra API call? Can't we test if a user can act as a proxy by checking the user object?

app/pages/select/routes.js
```diff
@@ -0,0 +1,21 @@
...
+  router.get('/', authProvider.authorise({ claim: 'ordering' }), withCatch(logger, authProvider, async (req, res) => {
+    const accessToken = extractAccessToken({ req, tokenType: 'access' });
+    const context = await getSelectContext({
+      accessToken,
+      orgId: req.user.primaryOrganisationId,
+      orgName: req.user.primaryOrganisationName,
+    });
+    logger.info('navigating to organisation selection page');
+    res.render('pages/select/template.njk', addContext({ context, user: req.user }));
+  }));
```
bjss-chris, line 18: Can you add some tests for the new route?

bjss-chris (changes requested):

app/helpers/api/oapi/getOrganisation.js
```diff
@@ -12,3 +12,10 @@ export const getOrganisation = async ({ orgId, accessToken }) => {
   logger.info(`Organisation with id: ${orgId} found in OAPI`);
   return organisationData;
 };
+
+export const getRelatedOrganisations = async ({ accessToken, orgId }) => {
+  const orgDataEndpoint = `${getOrganisationEndpoint({ orgId })}/related-organisations`;
+  const organisationData = await getData({ endpoint: orgDataEndpoint, accessToken, logger });
+  logger.info('Get Related Organisations found in OAPI');
+  return organisationData;
+};
```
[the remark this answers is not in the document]
KeithLyonMastek: There is a mismatch of this design pattern, about 50/50, some files have multiple export functions others just have 1 function per file.  I will create a separate file for the function to keep with your recommendation

app/pages/select/controller.js
KeithLyonMastek: What attribute of the user object has the setting that lets us know it is a proxy?

bjss-chris: The changes @leerogers-bjss is working on will add a `relatedOrganisationId` claim to the user if their primary organization has any related organizations. (It's essentially the same I what I referred to as `secondaryOrganisationId`.

This claim will only be present if related organizations exist, so I think we should be able to test for `req.user.relatedOrganisationId`.

KeithLyonMastek: This will need to be a separate ticket and we will implement that change when it has been finished

bjss-chris: Why does it need a new ticket?

KeithLyonMastek: This PR will have to sit open for x number of days until the change has been made.  We put in dummy values while we wait for api work to be finished, then we have a ticket to update the dummy values with real values when the api has been approved.  Otherwise we have a blocker for this PR

bjss-chris: The API changes are in PR and approved. However, even if the API changes weren't quite ready this could still go in (with the suggested change to check the claims) without any need for it to sit open.

KeithLyonMastek: please link the pr so i can look at the story are carry on with this ticket

app/pages/dashboard/controller.test.js
```diff
@@ -32,7 +33,9 @@ describe('dashboard controller', () => {
       await getDashboardContext({ orgId: 'org-id', orgName: 'org1', accessToken: 'access_token' });
 
       expect(contextCreator.getContext.mock.calls.length).toEqual(1);
-      expect(contextCreator.getContext).toHaveBeenCalledWith({ orgName: 'org1', completedOrders: [], incompletedOrders: [] });
+      expect(contextCreator.getContext).toHaveBeenCalledWith({
+        orgName: 'org1', completedOrders: [], incompletedOrders: [], userIsProxy: undefined,
```
bjss-chris, line 37: Can we mock `getIsUserProxy` to test that the expected value is passed through rather than `undefined`?

app/pages/select/routes.test.js
```diff
@@ -0,0 +1,46 @@
...
+  mockAuthorisedCookie,
+  setUpFakeApp,
+} from '../../test-utils/routesTestHelper';
+import * as controller from './controller';
+
+jest.mock('../../helpers/api/oapi/getRelatedOrganisations');
+
+describe('GET /organisation/select', () => {
+  const path = '/organisation/select';
+
```
bjss-chris, line 17: Probably worth adding an `afterEach` to reset all mocks – there may not be any issues now but it helps reduce the likelihood of any issues occurring with incorrect mocking in the future.

```js
afterEach(() => {
    jest.resetAllMocks();
  });
```

bjss-chris (commented):
Updates look good – just a couple of comments.

Also, @leerogers-bjss has merged the API claim PR so that will be available very shortly (if it isn't already).

app/pages/select/controller.js
bjss-chris: https://github.com/nhs-digital-gp-it-futures/BuyingCatalogueIdentity/pull/314

bjss-chris (approved):

PR #642, opened 2021-05-11
Feature/13275 fix bulk price validation

NickNewlands:


app/helpers/controllers/validateOrderItemFormBulk.js
```diff
@@ -36,7 +36,7 @@ export const validateOrderItemFormBulk = ({ data, selectedPrice, orderItemType }
         field: 'Price',
         id: 'PriceLessThanMax',
       });
-    } else if (parseFloat(data.price) > selectedPrice.price) {
+    } else if (parseFloat(data.price) > (selectedPrice.originalPrice || selectedPrice.price)) {
```
bjss-chris, line 39: Should this be `} else if (parseFloat(data.price) > selectedPrice.originalPrice) {`?

As we only want the validation error to occur if the entered price is greater than the list price.

app/helpers/controllers/validateOrderItemFormBulk.test.js
```diff
@@ -169,10 +169,10 @@ describe('validateOrderItemFormBulk', () => {
       expect(errors).toEqual([priceLessThanMax]);
     });
 
-    it('should return an array of one validation error if price is greater than the list price', () => {
+    it('should return an array of one validation error if price is greater than the list price and no orignal price exists', () => {
       getSelectedPriceManifest.getSelectedPriceManifest.mockReturnValue(selectedPriceManifest);
       const data = {
-        price: '0.11',
+        price: '1.1',
```
bjss-chris, line 175: As above, I'm not sure this is correct. If we don't have a list price we can't perform the validation but that should never occur as the order form won't work without an available list price.

app/helpers/controllers/validateOrderItemFormBulk.test.js
```diff
@@ -182,6 +182,46 @@ describe('validateOrderItemFormBulk', () => {
       expect(errors).toEqual([priceGreaterThanListPrice]);
     });
 
+    it('should not error if price is greater than the list price but less than the orignal price', () => {
+      getSelectedPriceManifest.getSelectedPriceManifest.mockReturnValue(selectedPriceManifest);
+      const data = {
```
bjss-chris, line 187: This is confusing – it should error if the entered price is greater than the list price. I'm not sure what original price is in this context.

There are only two prices that we should be concerned with: the entered price and the list price.

bjss-chris (changes requested):
Looks like the tests are failing – I'd hazard a guess that there's probably some mocked session that needs `priceId` adding to it. However, I think some of the naming used is confusing and I think there is probably some logic that needs tweaking too.

app/helpers/controllers/validateOrderItemFormBulk.js
NickNewlands: I just left in the existing logic in case it is use anywhere else an added the possibility of having an 'original price'.

PR #666, opened 2021-05-21
added cookie expiration using env variable

nehajain008:
Story [AB#7250](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/7250)
Task [AB#14642](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/14642)

### DIFF
--- a/app/config.js
+++ b/app/config.js
@@ -75,4 +75,7 @@ module.exports = {
 
   // content and user journey if they can order for multiple organisations
   showProxy: process.env.SHOW_PROXY || 'true',
+
+  // Cookie consent expiration date
+  buyingcatalogueCookieConsentExpiration: 1621594398846,
 };
--- a/app/config.js
+++ b/app/config.js
@@ -75,7 +75,4 @@ module.exports = {
 
   // content and user journey if they can order for multiple organisations
   showProxy: process.env.SHOW_PROXY || 'true',
-
-  // Cookie consent expiration date
-  buyingcatalogueCookieConsentExpiration: 1621594398846,
 };
--- a/app/routes.js
+++ b/app/routes.js
@@ -1,7 +1,7 @@
 import express from 'express';
 import {
   ErrorContext, errorHandler, healthRoutes, authenticationRoutes,
-  cookiePolicyAgreed, cookiePolicyExists, cookiePolicyClear,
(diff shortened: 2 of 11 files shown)


app/routes.js
```diff
@@ -30,13 +30,27 @@ const addContext = ({ context, req, csrfToken }) => ({
 
 export const routes = (authProvider, sessionManager) => {
   const router = express.Router();
+  const a = new Date();
```
kiran-mastek, line 33: Do we need to use milliseconds? 
The config value can be passed in as date for comparison.

Also do we not need tests for the code or have they been added in another check in?

app/config.js
```diff
@@ -75,4 +75,7 @@ module.exports = {
 
   // content and user journey if they can order for multiple organisations
   showProxy: process.env.SHOW_PROXY || 'true',
+
+  // Cookie consent expiration date
+  buyingcatalogueCookieConsentExpiration: 1621594398846,
 };
```
NickNewlands, line 81: This needs to be defined in buying-catalogue-library otherwise you would have to update this variable in every front-end project to expire the cookie.

app/routes.js
```diff
@@ -30,13 +30,27 @@ const addContext = ({ context, req, csrfToken }) => ({
...
+      const cookieName = 'buyingcatalogue-cookie-consent';
+      const value = req.cookies[cookieName] ? JSON.parse(req.cookies[cookieName]) : undefined;
+      if (value && value.creationDate
+        && (value.creationDate < config.buyingcatalogueCookieConsentExpiration)) {
+        cookiePolicyClear({ res, logger });
+      }
+      return next();
+    });
+  }
+
```
NickNewlands, line 53: You could move router.use into buying-catalogue-library just to keep all the logic together. The error message is done in a similar way if you want to see an example.

app/routes.js
NickNewlands: I don't think there is any problem in using a timestamp, it is very easy to get a hold of and compare and I think it avoids potential problem with timezones.

kiran-mastek: Timestamp is ok then but unit tests should be added

nehajain008:
Added suggested changes in another branch feature/7250-cookie-expirable-by-env-variable

